PR #5, opened 2026-08-13
Fix Phoenix pyproject license path

Pitchfork-and-Torch:
## Summary
`phoenix/pyproject.toml` had `license = { file = "LICENSE" }`, but there is no `phoenix/LICENSE`. The Apache-2.0 license lives at the repository root. PEP 621 `license.file` is relative to the pyproject, so packaging tools looking for `phoenix/LICENSE` fail.

This PR uses `license = { text = "Apache-2.0" }` (already in the SPDX header). README already points at the root LICENSE file.

## Test plan
- [ ] No `phoenix/LICENSE` required for `uv` / PEP 621 metadata
- [ ] SPDX header still Apache-2.0

### DIFF
--- a/phoenix/pyproject.toml
+++ b/phoenix/pyproject.toml
@@ -6,7 +6,7 @@ version = "0.1.0"
 description = "Phoenix — X's open-source recommendation algorithm (ranking + retrieval, single-host JAX)."
 readme = "README.md"
 requires-python = ">=3.11"
-license = { file = "LICENSE" }
+license = { text = "Apache-2.0" }
 dependencies = [
     "dm-haiku==0.0.16",
     "jax[cuda12]==0.8.1 ; sys_platform == 'linux'",
--- a/phoenix/pyproject.toml
+++ b/phoenix/pyproject.toml
@@ -6,7 +6,7 @@ version = "0.1.0"
 description = "Phoenix — X's open-source recommendation algorithm (ranking + retrieval, single-host JAX)."
 readme = "README.md"
 requires-python = ">=3.11"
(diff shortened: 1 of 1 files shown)


ting-hong-shieh (commented):
The missing-path diagnosis looks right. Could this use the current PEP 639 form instead: license = "Apache-2.0"? The proposed license = { text = "Apache-2.0" } is the legacy/deprecated table form and records the identifier as free-form license text rather than a License-Expression. PyPA guidance: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files

There is a second packaging detail worth verifying: removing the broken file reference fixes metadata loading, but does not by itself ensure that the full Apache license text is included in the wheel/sdist. license-files patterns are relative to phoenix/ and cannot use .., so the package may need a license file within its project root or a different packaging-root arrangement. Could the test plan include building the distribution and inspecting its METADATA and .dist-info/licenses contents?

Pitchfork-and-Torch:
Updated to the PEP 639 string form: license = "Apache-2.0".

On license-files: phoenix/ is the packaging root and PEP 639 patterns cannot walk to the repo-root LICENSE. This PR only unblocks metadata load from the broken file = "LICENSE" path. Shipping the full Apache text inside the wheel/sdist would need a license file under phoenix/ or a different packaging root; that is a separate change.

Public dump still cannot build the internal xai-* path deps, so I did not add a build-and-inspect-METADATA test here.

BAPPITTO (approved):

Pitchfork-and-Torch:

PR #6, opened 2026-08-13
Ignore macOS feature-switch test that needs internal YAML

Pitchfork-and-Torch:
## Summary
`test_feature_switches` is compiled on macOS and loads `/Users/$USER/workspace/config/features/home-mixer/main/*.yml`. Those files are not in this dump, so `cargo test` on a Mac fails immediately.

This PR marks the test `#[ignore]` with a comment. No ranking changes.

## Test plan
- [ ] `cargo test` on macOS no longer fails at FeatureSwitches::load_file
- [ ] Test remains in tree for xAI laptops that have the YAML

### DIFF
--- a/home-mixer/candidate_pipeline/phoenix_candidate_pipeline.rs
+++ b/home-mixer/candidate_pipeline/phoenix_candidate_pipeline.rs
@@ -1010,6 +1010,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore = "requires internal xAI feature-switch YAML not shipped in this repo"]
     #[cfg(target_os = "macos")]
     fn test_feature_switches() {
         use std::env;


jofjac2-hue:
What characteristics does the system appear to reward when deciding which posts reach people who don't already follow the artist?

Pitchfork-and-Torch:
@jofjac2-hue This PR only `#[ignore]`s a macOS test that needs internal YAML, so it does not change distribution. The published For You path does answer the question, though.

Posts that reach people who do not already follow the author take a different *retrieval* door than follower posts, then the same ranker with an out-of-network haircut.

**How a stranger's post gets considered at all**

In-network is Thunder (recent posts from accounts the viewer follows). Out-of-network is Phoenix retrieval plus SimClusters (README Overview).

- Phoenix retrieval is a two-tower model. The viewer is encoded from *what they recently engaged with* (no learned per-user ID). Candidates are semantic IDs from multimodal embeddings plus hashed author IDs. Retrieval is nearest-neighbor to that history (`phoenix/README.md`).
- SimClusters seeds from the viewer's recent explicit and implicit engagements and pulls other posts by cosine similarity of log-fav tweet embeddings (`home-mixer/sources/simclusters_source.rs`). Score floor 0.5, age window 48h.

The first characteristic is not "globally popular." It is *looks like what this viewer already engaged with*.

**What the ranker then rewards**

Phoenix predicts P(action) per viewer and post. `RankingScorer` does a weighted sum. Defaults live in `home-mixer/params/param.rs`:

Positive (high to low):
- share via copy link: 20
- reply: 5, quote: 5, share via DM: 5
- follow author: 4
- share: 2
- retweet: 1
- favorite: 0.5
- click: 0.4, open link: 0.2
- photo expand / video open / video quality view: 0.05 each
- dwell time (continuous): 0.004
- profile click: 0.0, binary dwell: 0.0

Negative:
- report: -234
- mute author: -58.8
- not interested: -43.2
- block author: -31.2
- not dwelled: -0.02

For non-followers, the bidirectional-follow reply boost (+15, so reply becomes 20 for mutual follows) does **not** apply. `PostUnexploredWeight` is in-network only by default.

After the sum:
- OON multiplier 0.75 (`OonWeightFactor`; 0.5 on topic feeds)
- author diversity: extra posts from the same author decay by 0.5 down to a 0.25 floor
- cold-start lift: original posts from authors with <=1000 followers and <1000 impressions, younger than 24h, can be lifted toward a target slot
- `VMRanker` then trades a bit of score for less embedding-similar neighbors (DPP)

**Hard gates for non-followers (not rewards, but they decide reach)**

- `OONRetweetReplyFilter`: replies and reposts from accounts the viewer does not follow are dropped. Original posts (and quotes that are not replies or RTs) can stay.
- `AgeFilter`: older than 48 hours is gone.
- Visibility filtering is stricter OON: spam high-recall, NSFW author, NSFW text, malicious URL, geo-restricted media, FosnrAbuseInsults, and similar can DROP for non-followers while ALLOW for followers (`visibility-filtering/rules/registry.rs`).
- Protected authors do not reach non-followers.
- Viewer blocks, mutes, muted keywords, and already-seen still apply.

**Plain reading**

The dump does not score "is this good art." It scores "will *this* viewer reply, quote, follow, share, or linger, versus mute, block, or report." Getting into the OON pool means matching the viewer's recent engagement neighborhood (Phoenix SIDs + SimClusters log-fav). Surviving there means an original, recent post that the model thinks will draw high-intent actions without predicted negative feedback, after a 25% OON discount and author-diversity decay.

BAPPITTO (approved):

chaosnull146-arch (changes requested):
Should add unit for changes

Pitchfork-and-Torch:
@chaosnull146-arch This PR only `#[ignore]`s `feature_switch` on macOS because that test reads internal YAML that is not in the public dump. It does not change ranking or serving behavior.

There is no new unit to add without that YAML (or a stub that would no longer test the real loader). Happy to add a compile-only note in the ignore reason if that is what you meant.

Pitchfork-and-Torch:

PR #7, opened 2026-08-13
Fail closed when safety hydration fails

navigotechsolutions-labs:
## Summary

The visibility-filtering hydration layer converts author and viewer-relationship backend failures into default or empty metadata. Downstream rules can then evaluate the candidate as ordinary content and return `Allow`, even though safety-critical metadata was unavailable.

## Fix

- Track failed author and social-graph hydration per candidate.
- Preserve the candidate while marking safety-critical hydration failure.
- Add a first-position visibility rule that drops marked candidates before content-specific rules can allow or interstitial them.

## Tests

- Add hydration assembly coverage for failed author hydration.
- Add policy coverage proving failed hydration drops on both home surfaces.
- Add coverage proving healthy candidates remain eligible.
- `git diff --check` passes.

The repository checkout does not expose a Cargo manifest for `visibility-filtering`, and the environment lacks `cargo`, `rustc`, and `rustfmt`, so native Rust tests and formatting could not be run locally. Source-level call-site and struct-literal audits were completed.

### DIFF
--- a/visibility-filtering/hydration/batch.rs
+++ b/visibility-filtering/hydration/batch.rs
@@ -114,6 +114,10 @@ impl<K: Eq + Hash, V> HydrationBatch<K, V> {
         self.results.values().filter(|r| r.is_failed()).count()
     }
 
+    pub(crate) fn is_failed(&self, key: &K) -> bool {
+        self.results.get(key).is_some_and(Hydrated::is_failed)
+    }
+
     pub(crate) fn hydrated(&self, key: &K) -> Option<&Hydrated<V>> {
         self.results.get(key)
     }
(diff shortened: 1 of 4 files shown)


Pitchfork-and-Torch (commented):
## Review (COMMENT, not approve)

The fail-closed rule on failed author or social-graph hydration is the right serving default. Empty/default features on those paths currently look like a healthy user with no blocks, which is fail-open.

Gap still open after this PR: `safety_labels` still default to an empty map when hydration misses (`unwrap_or_default` in assemble). That remains fail-open for restriction labels. Pair with PR 12 (expiry at serve) rather than treating empty labels as no restrictions.

`FilteredReason::UnspecifiedReason` will be opaque in Under the Hood / debug. A dedicated reason would make the drop inspectable.

Tests for both home surfaces and the healthy-candidate still-Allow case are the right shape. Public dump still has no standalone Cargo target for this crate, so I did not rerun rustc.

Verdict: support-with-gap. I would take this, then fail-close or explicitly metric safety-label hydration misses.

BAPPITTO (approved):

PR #9, opened 2026-08-13
Fix muted authors resurfacing through retweets

navigotechsolutions-labs:
## Summary

`AuthorSocialgraphFilter` checked the viewer’s block list for `retweeted_user_id` but did not apply the viewer’s mute list. As a result, a post from a muted original author could resurface when another account retweeted it.

## Fix

Apply `viewer_muted_user_ids` to `candidate.retweeted_user_id` and remove the candidate when the original retweeted author is muted.

## Tests

Added regression coverage for both a muted retweeted author (removed) and an unmuted retweeted author (kept).

`git diff --check` passes. The full Rust test suite could not be run in the local environment because this checkout does not expose a `home-mixer` Cargo manifest and the environment lacks `cargo`, `rustc`, and `rustfmt`.

### DIFF
--- a/home-mixer/filters/author_socialgraph_filter.rs
+++ b/home-mixer/filters/author_socialgraph_filter.rs
@@ -41,6 +41,10 @@ impl Filter<ScoredPostsQuery, PostCandidate> for AuthorSocialgraphFilter {
                 .retweeted_user_id
                 .map(|uid| viewer_blocked_user_ids.contains(&(uid as i64)))
                 .unwrap_or(false);
+            let viewer_mutes_retweeted_user = candidate
+                .retweeted_user_id
+                .map(|uid| viewer_muted_user_ids.contains(&(uid as i64)))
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch (commented):
## Review (COMMENT, not approve)

Applying `viewer_muted_user_ids` to `retweeted_user_id` matches the existing block check on the same field. Muted original authors should not re-enter For You via a third-party retweet.

Same asymmetry is still open for quotes: `viewer_blocks_quoted_author` exists, but there is no `viewer_mutes_quoted_author` on `quoted_user_id`. If you are already in this filter, adding mute-on-quote in the same PR (plus one test) would close the pair.

Regression tests for muted vs unmuted retweeted author are sufficient for this hole.

Verdict: support-with-gap. Take the retweet mute; please also cover quoted authors or say why quotes are intentionally different.

PR #16, opened 2026-08-13
Register BroadcastLivenessHydrator module

Pitchfork-and-Torch:
## Summary
`broadcast_liveness_hydrator.rs` exists (with tests) and `docs/BIDIRECTIONAL_BOOST_CHANGE.md` shows the `pub mod` / use lines, but `candidate_hydrators/mod.rs` never declared the module, so the hydrator and its tests never compiled.

This PR only adds the `pub mod` line in alpha order. It does not insert the hydrator into a pipeline.

## Test plan
- [ ] `pub mod broadcast_liveness_hydrator` is in `home-mixer/candidate_hydrators/mod.rs`
- [ ] No new pipeline wiring


Pitchfork-and-Torch:

PR #20, opened 2026-08-13
Backfill candidates after late filtering

cmdr-chara:
## Summary
- add opt-in same-request reserve backfill after late filtering
- enable it only for Phoenix, whose TopK selector guarantees real ranked reserve candidates
- incrementally hydrate only the number needed and preserve whole-slate conversation deduplication
- preserve exact legacy behavior for generic/custom selectors whose `non_selected` records may be synthetic side-effect placeholders

## Verification
- tests for >15 late drops, multi-round refill, ordering, cross-batch deduplication, and candidate conservation
- regression proving default/custom pipelines never hydrate or promote synthetic placeholders
- independent post-implementation audit approved the corrected two-commit branch
- `git diff --check` / `git show --check`

The published candidate-pipeline subtree has no Cargo/BUILD manifest, so tests could not be executed locally.

### DIFF
--- a/candidate-pipeline/candidate_pipeline.rs
+++ b/candidate-pipeline/candidate_pipeline.rs
@@ -90,6 +90,13 @@ where
     fn post_selection_filters(&self) -> &[Box<dyn Filter<Q, C>>];
     fn side_effects(&self) -> Arc<Vec<Box<dyn SideEffect<Q, C>>>>;
     fn result_size(&self) -> usize;
+    /// Whether `Selector::non_selected` is a ranked reserve that is safe to hydrate and serve.
+    ///
+    /// Custom selectors may use `non_selected` for side-effect accounting or placeholders, so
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch (commented):
Leftover serving hole after this class.

Phoenix now opts in (enable_post_selection_backfill -> true) so late VF / conversation dedup can refill from the selector reserve. ReverseChronPostsPipeline still runs VFFilter + AncillaryVFFilter after select (reverse_chron_posts_pipeline.rs) and inherits the default false.

Viewer effect: Following reverse-chron can return a short slate after late VF even when unused ranked reserve exists. Twin: the Phoenix override in this PR.

Following / ranked-following post-select filter vecs are empty, so they are not the same hole. For You blender only runs AdAdjacentServedFilter, which is a different class.

Not blocking the Phoenix opt-in. A follow-up that enables backfill on ReverseChron (or proves PassthroughSelector has no serveable reserve) would close the family.

PR #31, opened 2026-08-14
Add meritocratic author-size IPS to RankingScorer

Pitchfork-and-Torch:
## Summary

Phoenix already estimates P(action | viewer, post). That is the quality signal. Follower count is not a Phoenix feature, but audience size still leaks into For You through Thunder (follow graph), hashed author IDs, SimClusters log-fav retrieval, and the flat 0.75 OON tax.

This PR residualizes ln(1 + followers) inside the scored batch so two posts with the same Phoenix score are not ordered by how many people already follow the author.

```
p_i   = ln(1 + max(followers_i, 1))
ips_i = 1 / p_i
m_i   = 1 + alpha * (ips_i / mean(ips) - 1)
m_i   = clamp(m_i, 1 / max_boost, max_boost)
score' = score * m    if score > 0
       = score        otherwise
```

Mean-normalization keeps the batch arithmetic mean at 1. This reallocates score. It does not inflate it. Missing follower counts are identity. Negative scores (predicted block / mute / report) are not lifted.

Quality still wins. Default clamp is 2x, so a 1M-follower post that Phoenix scored 3x higher still ranks above a 100-follower post.

This is individual meritocratic fairness (Singh and Joachims KDD 2018, groups of size one; Biega et al. equity of attention). It is not demographic parity. No identity groups. No impression quotas. Low-quality posts stay low.

## Why here, not UserCred / retrieval

- UserCred is an enforcement prestige graph. Ranking does not read it. Prestige-as-immunity is already in #18 / #19.
- SidTail (authors under 1k followers) is the retrieval counterpart. A post that never enters the slate cannot be residualized. This PR does not invent a new cluster.
- Cold start (#22) lifts one low-impression original into slot ~15. IPS applies to every scored post, continuously.
- Author diversity only decays the 2nd+ post from the same author.

## Knobs

| Param | Default |
|---|---|
| rust_home_mixer_enable_author_size_ips | true |
| rust_home_mixer_author_size_ips_alpha | 0.5 |
| rust_home_mixer_author_size_ips_max_boost | 2.0 |

`alpha = 0` is a no-op. Applied after the Phoenix weighted sum, before author diversity, the OON tax, and cold start.

Math writeup: docs/MERITOCRATIC_AUTHOR_SIZE_IPS.md

## Test plan
- [x] identity when alpha=0, followers missing, or all authors the same size
- [x] smaller author gets a higher multiplier; known-multiplier mean is 1
- [x] clamp respects max_boost; 0 followers treated as 1
- [x] negative / zero scores are not boosted
- [x] equal Phoenix scores: 100-follower ranks above 1M-follower
- [x] 3x Phoenix gap still beats the default 2x clamp
- [x] RankingScorer integration: enable lifts, disable restores equality
- [ ] cargo test cannot run in the public dump (no home-mixer manifest)

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,8 +342,9 @@ Positive actions carry positive weights, negative actions negative ones. The wei
 
 There is a common misconception to be aware of about the weights: they scale the predicted probabilities (or predicted continuous values, e.g. dwell time) — they do *not* scale the raw engagement counts, so e.g. it'd be incorrect to see that a report has 468 times higher weight than a like and conclude that e.g. "1 report cancels out 468 likes". The weights are a multiple on your own predicted probability of Liking, Reporting, etc, which is substantially driven by your own behavior.
 
-Three adjustments follow:
+Four adjustments follow:
 
+- **Author-size IPS**: a batch-mean-normalized inverse-propensity multiplier on `ln(1 + followers)`. Equal Phoenix scores are not ranked by audience size. Quality still wins: the default clamp is 2x, so a large-author post that is 3x more relevant still ranks higher. Math: [`docs/MERITOCRATIC_AUTHOR_SIZE_IPS.md`](docs/MERITOCRATIC_AUTHOR_SIZE_IPS.md).
 - **Author Diversity**: each post after an author's first is multiplied by a decaying factor, down to a floor.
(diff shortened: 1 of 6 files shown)


Pitchfork-and-Torch:

Issue #25, opened 2026-08-13
BDSM enforcement policy is undeployable: redacted thresholds [9.99] lack configuration guide


vishalsingh2972:
## Problem
The BDSM (Behavioral Inauthentic-Account Detection) module's safety thresholds 
are redacted in the public release to prevent adversarial evasion, but the 
documentation and code lack clear guidance on how to create custom enforcement policies.

This undermines the repository's transparency mission, as users cannot understand:
- What threshold values are expected
- How to configure their own operating points
- The relationship between sentinel `9.99` and valid threshold ranges

## Current State

**Redacted threshold locations:**
- `bdsm/runtime/sink_policy.yaml` (lines 8-31): All thresholds set to `9.99` sentinel
- `bdsm/runtime/score_results_sink_focal.py` (lines 175-207): Default thresholds all set to `9.99`

**Example of redaction:**
```yaml
thresholds:
  FollowBot:           [9.99, 9.99]
  EngagementAmplifier: [9.99, 9.99]
  RTBot:               [9.99, 9.99]
```

All probabilities in these fields should be in the range `[0, 1]`, making `9.99` 
an obvious placeholder (from `bdsm/README.md` lines 104-112).

**Documentation issue:**
The README states users should "Supply your own via `--policy-file` / `BDSM_SINK_POLICY`" 
but provides no template or example showing:
- The expected structure of each threshold (2-tuple: `[tau, lambda]`)
- Valid value ranges (`[0, 1]` for probabilities)
- Which fields are safe to modify (vs. which control system behavior)

## Impact
Users attempting to run BDSM cannot deploy it without either:
1. Reverse-engineering the policy structure from code parsing
2. Providing arbitrary `9.99` values that will never trigger enforcement
3. Guessing appropriate thresholds without domain knowledge

This is particularly problematic for the stated goal of the repo: **public transparency 
into how the For You feed works**.

## Proposed Solution

1. **Create example template:** `bdsm/runtime/sink_policy.example.yaml`
   - Show realistic threshold values in `[0, 1]`
   - Comment each field explaining its purpose
   - Label which values are likely to be too conservative/aggressive

2. **Update `bdsm/README.md`:**
   - Document the policy file format with concrete examples
   - Explain the 2-tuple structure: `[tau, lambda]` as threshold and ceiling
   - Add a "Configuration" section showing how to set policy via CLI or env var
   - Link to the example file

3. **Add validation at startup:**
[...]

PR #35, opened 2026-08-14
Validate redacted score-results sink policies

ting-hong-shieh:
## Summary

- document the results-sink policy pair semantics and the public 9.99 redaction sentinel
- validate policy tables, head names, numeric finiteness, ranges, and sentinel consistency at load time
- warn when a policy contains redacted fields and explicitly disable affected lanes
- add focused tests for valid, malformed, and redacted policies

## Why

The checked-in public policy uses 9.99 to represent redacted thresholds. The cusp lane previously evaluated tau - cusp_delta; with the public values, 9.99 - 9.99 becomes 0. That arithmetic meant the sentinel did not reliably keep the lane disabled.

This change gives the sentinel explicit runtime behavior instead of relying on out-of-range comparisons. It does not propose, infer, or expose production threshold values.

## Validation

- 42 passed, 3 skipped in bdsm/tests
- Ruff format check passed
- Ruff core error and import checks passed
- git diff --check passed

Addresses #25

### DIFF
--- a/bdsm/README.md
+++ b/bdsm/README.md
@@ -124,3 +124,37 @@ The scorer publishes an 8-wide row in `heads.HEAD_NAMES` order.
   it carries that sentinel plus the model-head suffix, not the internal
   appeal paragraph or the action types each head keys off. ActionName
   proto enums are unchanged.
+
+### Results-sink policy format
+
+Pass a reviewed policy with `--policy-file PATH`, or set `BDSM_SINK_POLICY` to
+that path. The loader falls back to `runtime/sink_policy.yaml` when neither is
+set.
+
+Each entry in `thresholds`, `cusp_heads`, `paused_liveness_thresholds`, and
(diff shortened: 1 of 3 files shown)


valorjesson-commits (approved):

valorjesson-commits (approved):

valorjesson-commits (approved):

valorjesson-commits (approved):

ting-hong-shieh:
Rebased onto `b089ce6` to clear the conflict with `main`.

The added and removed lines are identical to the reviewed `becb6b7`. `main` had independently added `build_enforcement_note()` and two enforcement-note tests in the same regions of `score_results_sink_focal.py`, `test_sink_policy.py`, and `README.md`; the resolution keeps both sides, so this PR's payload is unchanged.

Validation on `ce312de`:

```
44 passed, 3 skipped   # bdsm suite
```

PR #52, opened 2026-08-14
Carry coarse label-source provenance into Under the Hood reports

### LINKED ISSUE #39: Under the Hood does not carry label-source provenance into the report, contrary to README:411
README:411 lists as a benefit of Under the Hood:

> one can see whether labels have been manually applied outside of automated systems

I could not find a path in the released code that carries per-event label source into the report.

**Source is structured where the label is read.** `visibility-filtering-client/tweet_safety_label.rs:197-209` maps `safety_label_source` into three distinct variants: `BotmakerAction` (carrying `rule_id`), `ToolAction` (carrying `agent_tool` and `actor_ldap`), and `GrokAnnotationAction` (carrying `source`). `xai_x_thrift` is not published, so I am not claiming what each variant denotes. The point is only that a per-event source discriminator exists and is structurally distinct at this layer.

**It is not carried through the Under the Hood path.**

- `under-the-hood/scalding/UthDailyPostsJob.scala:263` retains `event.labelType.originalName`. The event body is read at `:266` and `:269` for `createdAtMsec` and `expiresAtMsec` only.
- `under-the-hood/thrift/uth_serving.thrift:53-61`: `UthPostLabelAggregate` and `UthAccountLabelAggregate` hold a label name and day counts. There is no source field to populate.
- `under-the-hood/strato/columns/underTheHoodReport.User.strato:194-199` and `:210-215` emit label, about, effect, and counts.
- `under-the-hood/strato/graphql/under_the_hood.thrift:9-13`: the response is `generatedAt`, `reportJson`, `reportInfo`.

**The static `about` text does not stand in for it.** Those strings are per-label-type constants, and seven of them state the disjunction rather than resolving it. `underTheHoodLabels.strato:12` reads "Post detected by automated systems or in response to a user report", with the same construction at `:30`, `:40`, `:61`, `:77`, `:95`, and `:120`.

Two resolutions look legitimate to me:

1. Carry a coarse category derived from the variant, for example automated, manual, or LLM, without exposing `rule_id` or `actor_ldap`.
2. Amend README:411 to describe what the report contains.

I searched the tracker for `manually applied`, `actor_ldap`, `ToolAction`, `BotmakerAction`, `GrokAnnotation`, `label source`, and `source provenance`, with no matches. #11 edits two of these files and its description notes it avoids inventing source metadata absent from the aggregate. If you consider that the same target, say so and I will fold this in rather than track it separately.

Audited at `a389166f6cf5da70a286b568c87695d4dcdce3a1`. This is a static reading of the checked-in types and job code. `under-the-hood/` and `visibility-filtering-client/` ship no build target, so I have not executed anything.

Pitchfork-and-Torch:
## Summary

Fixes https://github.com/xai-org/x-algorithm/issues/39.

README says Under the Hood should show whether labels were manually applied outside automated systems. The daily job already reads `TweetSafetyLabelEvent.label` but kept only timestamps; `SafetyLabelSource` was dropped, and the serving types had no field to carry it.

`reportJson.source` is one of four tokens:

- `automated` ← `BotMakerAction`
- `manual` ← `ToolAction`
- `llm` ← a Grok/LLM union member, detected by `productPrefix` or `getName` plus the last non-empty `.` / `$` segment (`GrokAnnotation*`), skipping a trailing synthetic `$` and `getSimpleName` (can throw on Scala `$` names), so this compiles if that case is absent from `spam.rtf`
- `unknown` ← unset snapshot / tweet-flag / unmapped-source rows (serving clamp)

Jobs persist only `automated` | `manual` | `llm` after trim/lowercase. Unmapped, missing, leftover tokens, and tweet-flag rows stay unset (`source = None` is set explicitly on flag writes). Daily counts clamp the last-apply source before grouping. When an event apply and an unset snapshot apply are merged, the later timestamp wins; a persistable event source is kept only if the later row is a snapshot gap-fill (`isSnapshot`), not a later unmapped event. `latestAsOf` treats the newest asOf as a full rewrite of `(user, authored day, label)`: it keeps every persistable source from that asOf and drops older asOf rows so a later sourced rewrite cannot sit beside an earlier unset row. The month MH writer reconstructs rows with named Scrooge fields and pattern-matches leftover same-day `carried` / `removed` (not `best._4` / `best._5`). Serving clamps any other stored string to `unknown`, merges leftover rows that clamp to the same `(label, source)` by unioning days (larger carried/removed per day, then sum; `groupBy.toList` before `flatMap` so Map.flatMap cannot drop rows), pattern-matches that leftover group key as `(name, source)` rather than `t._1` / `t._2`, and sorts that list by label then source so `reportJson` order is stable.

`postLabelSource` is a closed function (match + function closer before `formatPercentage`). Post-label `about` / `effect` look up by the raw label id, same as the account path; display `name` is only written to `label`.

Does not emit `rule_id`, `actor_ldap`, `agent_tool`, or VF-client type names. Snapshot and tweet-flag inputs stay unset (no invented source).

Based on current `xai-org/x-algorithm` main (`c65aa179`). Lane is `under-the-hood/` only.

## Test plan

- Static review of the event → daily → month → `reportJson` path.
- `under-the-hood/` has no published build/test target in this snapshot; no job was executed.
- Confirm Scrooge optional field `source` (ids 3 / 10) is backward-compatible with existing parquet/MH rows.
- Confirm `reportJson.source` can only be `automated` | `manual` | `llm` | `unknown`.
- Confirm snapshot/flag/unmapped/leftover-token rows coalesce to a single `unknown` row per label, including after daily count, `latestAsOf`, and reportJson day-union.
- Confirm a later unset snapshot apply does not wipe an earlier persistable event source.
- Confirm a later unmapped event apply does not inherit an earlier persistable source.
- Confirm a newer asOf that fills source replaces older unset rows for the same `(user, day, label)` and does not emit both.
- Confirm same-asOf `automated` and `manual` rows for one label both survive `latestAsOf`.
- Confirm tweet-flag daily/backfill rows set `source = None`.
- Confirm `Automated` / padded tokens normalize to the allowlist instead of `unknown`.
- Confirm post-label rows are ordered by `(label, source)`.
- Confirm leftover serving `groupBy` keys `(name, source)` by pattern match, not `t._1` / `t._2`.
- Confirm leftover month same-day counts use pattern-matched `carried` / `removed`, not `best._4` / `best._5`.
- Confirm llm matching uses `productPrefix` or `getName` plus the last non-empty `.` / `$` segment, not `getSimpleName` or `String.split("$")`.
- Confirm `postLabelSource` closes before `formatPercentage`.
- Confirm `postLabelAbout` / `postLabelEffect` receive the raw label id, not the display name.

### DIFF
--- a/under-the-hood/scalding/UnderTheHoodCommon.scala
+++ b/under-the-hood/scalding/UnderTheHoodCommon.scala
@@ -96,6 +96,9 @@ object UnderTheHoodCommon {
   def finalThroughDay(completeThroughDay: Int, asOfDay: Int, postObservationDays: Int): Int =
     math.min(completeThroughDay, addCalendarDays(asOfDay, -postObservationDays))
 
+  // Latest asOf is a full rewrite of (user, authored day, label). Keep every
+  // persistable source from that asOf; drop older asOf rows so a later sourced
+  // rewrite cannot sit beside an earlier unset row.
   def latestAsOfPostLabelRows(
     rows: TypedPipe[UthDailyPostLabel],
     reducers: Int
@@ -103,10 +106,36 @@ object UnderTheHoodCommon {
     applyReducers(
       rows.groupBy(r => (r.userId, r.authoredYyyymmdd, r.label)),
       reducers
-    ).reduce { (a, b) =>
-      if (a.asOfYyyymmdd.getOrElse(Int.MinValue) >= b.asOfYyyymmdd.getOrElse(Int.MinValue)) a
-      else b
-    }.values
+    ).mapValueStream { values =>
+      val all = values.toSeq
+      if (all.isEmpty) Iterator.empty
+      else {
+        val latest = all.map(_.asOfYyyymmdd.getOrElse(Int.MinValue)).max
+        all
+          .filter(_.asOfYyyymmdd.getOrElse(Int.MinValue) == latest)
+          .groupBy(r => UthLabelSource.persistToken(r.source))
+          .values
+          .iterator
+          .map { sameSource =>
+            val r = sameSource.maxBy { x =>
+              (x.carried.getOrElse(0L), x.removed.getOrElse(0L))
+            }
+            UthDailyPostLabel(
+              userId = r.userId,
+              authoredYyyymmdd = r.authoredYyyymmdd,
+              label = r.label,
+              carried = r.carried,
+              removed = r.removed,
+              asOfYyyymmdd = r.asOfYyyymmdd,
+              observationAgeDays = r.observationAgeDays,
+              isFinal = r.isFinal,
+              postObservationDays = r.postObservationDays,
+              source = UthLabelSource.persistToken(r.source)
+            )
+          }
+      }
+    }.toTypedPipe
+      .map { case (_, row) => row }
 
   def parseUserIds(args: Args): Set[Long] = {
     val raw = args.list("userIds").flatMap(_.split(",")).map(_.trim).filter(_.nonEmpty)
--- a/under-the-hood/scalding/UthDailyPostsBackfillJob.scala
+++ b/under-the-hood/scalding/UthDailyPostsBackfillJob.scala
@@ -100,7 +100,7 @@ class UthDailyPostsBackfillApp {
         observationMs,
         config)
         .map {
-          case (userId, authoredDay, label, asOfDay, carried, removed) =>
+          case (userId, authoredDay, label, source, asOfDay, carried, removed) =>
             val age = calendarDaysBetween(authoredDay, asOfDay)
             UthDailyPostLabel(
               userId = Some(userId),
@@ -111,7 +111,8 @@ class UthDailyPostsBackfillApp {
               asOfYyyymmdd = Some(asOfDay),
               observationAgeDays = Some(age),
               isFinal = Some(age >= config.postObservationDays),
-              postObservationDays = Some(config.postObservationDays)
+              postObservationDays = Some(config.postObservationDays),
+              source = UthLabelSource.persistToken(source)
             )
         }
 
@@ -135,7 +136,8 @@ class UthDailyPostsBackfillApp {
               asOfYyyymmdd = Some(authoredDay),
               observationAgeDays = Some(0),
               isFinal = Some(true),
-              postObservationDays = Some(config.postObservationDays)
+              postObservationDays = Some(config.postObservationDays),
+              source = None
             )
         }
 
@@ -204,10 +206,11 @@ object UthDailyPostsBackfillApp {
           if (expiresStr == null || expiresStr.isEmpty) Long.MaxValue else expiresStr.toLong
         if (expires <= rangeLookbackStartMs) None
         else if (createdStr == null || createdStr.isEmpty)
-          Some((tweetId, (labelName, Long.MinValue, true, expires)))
+          Some((tweetId, (labelName, Long.MinValue, true, expires, None, true)))
         else {
           val created = createdStr.toLong
-          if (created < rangeEndMs) Some((tweetId, (labelName, created, true, expires)))
+          if (created < rangeEndMs)
+            Some((tweetId, (labelName, created, true, expires, None, true)))
           else None
         }
     }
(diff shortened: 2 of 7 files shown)


kiri234-code:
Thank you for carrying coarse label-source provenance through the Under the Hood reporting path. This appears to address the reporting gap discussed in Issue #39.

I would like to clarify one important point about the scope of this change.

I have a real-world case where a post for which I voluntarily applied a content warning before publication appears in my Under the Hood report as `NSFW_HIGH_PRECISION`. The report shows one such post out of 73 total posts, with no account-level labels currently present.

This PR adds `source` to `UthDailyPostLabel` and `UthPostLabelAggregate`, but I do not see a corresponding change to `safety-label-user-agg`. Am I correct that this PR changes only how post-label provenance is reported, and does not change how those post labels are counted toward account-level NSFW labels?

If account aggregation evaluates `NSFW_HIGH_PRECISION` by label type without considering its source, an appropriately author-labeled post could still contribute to an account-level restriction in the same way as content that was left unlabeled and subsequently detected or reported.

Could you please clarify:

1. Whether an author-applied content warning is represented as `ToolAction` / `manual`, or whether it reaches this reporting path without a mapped `SafetyLabelSource` and therefore appears as `unknown`.
2. Whether `safety-label-user-agg` distinguishes author-applied labels from automated detections, reports, or X manual enforcement.
3. If it does not, whether preserving an `author_applied` category—or otherwise making source information available to account aggregation—should be addressed in this PR or tracked separately.

The provenance added by this PR is valuable for transparency, but the most important practical question for affected users is whether responsible self-labeling can itself contribute to account-level visibility restrictions.

Pitchfork-and-Torch:
@kiri234-code You are reading the scope correctly.

This PR only changes the Under the Hood reporting path (`under-the-hood/`). It adds a coarse `source` on `UthDailyPostLabel` / `UthPostLabelAggregate` and the reportJson tokens `automated` | `manual` | `llm` | `unknown`. It does not edit `safety-label-user-agg/` and it does not change how post labels are counted toward account-level labels.

On the three questions, from the published dump:

1. Author-applied content warnings are not a separate `author_applied` token. UTH maps `SafetyLabelSource.ToolAction` to `manual`, `BotMakerAction` to `automated`, and a Grok/LLM union member (`GrokAnnotation*`) to `llm`. Snapshot, tweet-flag, unmapped, and missing sources stay unset and clamp to `unknown` in the report. Whether a pre-publish content warning is stored as `ToolAction` vs a tweet-flag / unmapped snapshot is not proven in this dump, so a self-labeled post can show up as `manual` or `unknown` depending on how that event was written.

2. `safety-label-user-agg` does not distinguish source. The processor fetches `baseTweetSafetyLabelMap` and keeps `labels.keys` (label types only). Default rules trigger on `NSFW_HIGH_PRECISION` / `SOFT_NSFW` by name, then count matching posts in a window. No `SafetyLabelSource` check.

3. Making source available to account aggregation (or an `author_applied` category) should be a separate PR. This one is the reporting gap in issue #39. Expanding it into serving/agg would mix classes.

One post labeled `NSFW_HIGH_PRECISION` with no account-level labels is consistent with the default agg windows (account `NSFW_HIGH_PRECISION` wants 3 matching posts in a 5-post window). UTH can still show that one post.

xAI owns whether self-labeling should be excluded from account restrictions. This PR does not claim to change that.

kiri234-code:
@Pitchfork-and-Torch 
Thank you very much for the detailed and clear explanation.

This confirms the distinction I wanted to understand: this PR affects only the Under the Hood reporting path, while the published account aggregation logic currently evaluates label types without checking their source.

I also understand that the published code does not establish whether an author-applied warning is written as `ToolAction` / `manual`, or enters through a tweet flag or unmapped path and becomes `unknown`. Therefore, whether self-labeling produces the relevant post label remains a separate unresolved question.

Your explanation of the default 3-in-5 aggregation window also clarifies why one `NSFW_HIGH_PRECISION` post can appear in Under the Hood without producing an account-level label.

This answers my questions about the scope of the PR and the behavior visible in the published code. I appreciate you taking the time to examine each point so carefully.

Pitchfork-and-Torch:

PR #53, opened 2026-08-14
Fix Brazil election filter excluding a listed viewer's own id

Pitchfork-and-Torch:
## Bug

`Brazil2026ElectionFilter` treats a user as excluded when they are on the Electoral Court list **and** not in `query.user_features.followed_user_ids`.

That follow set is the social-graph following list. It never contains the viewer. So if the viewer is themselves a listed account, their own id is excluded.

That drops For You candidates that only surface the viewer:

- someone else retweets the viewer (`retweeted_user_id == query.user_id`)
- someone else quotes the viewer (`quoted_user_id == query.user_id`)
- a reply whose `ancestor_users` includes the viewer

`SelfTweetFilter` already removes the viewer's own posts. This hole is the amplification / thread-context path.

`SelfReplyChainFilter` already inserts `query.user_id` into its allowed set for the same reason. This filter did not.

Existing tests used `ScoredPostsQuery::default()` (empty follows, `user_id == 0`) and never covered the follower exception or self-id.

## Fix

Insert `query.user_id` into the allowed set before `should_remove`.

`is_excluded_author` is unchanged: listed **and** `!followed_user_ids.contains`. Unfollowed listed authors, retweets, quotes, and ancestors still drop.

## Tests

- listed author the viewer follows is kept; a different unfollowed listed author still drops
- retweet / quote / listed ancestor of the **viewer** are kept
- the same three shapes still drop when the listed id is not the viewer and is not followed
- reply is kept when the viewer follows the listed ancestor

Lane: `home-mixer/filters/brazil_2026_election_filter.rs` only. Based on `xai-org/x-algorithm` main (`c65aa179`).

### DIFF
--- a/home-mixer/filters/brazil_2026_election_filter.rs
+++ b/home-mixer/filters/brazil_2026_election_filter.rs
@@ -1383,12 +1383,16 @@ impl Filter<ScoredPostsQuery, PostCandidate> for Brazil2026ElectionFilter {
         query: &ScoredPostsQuery,
         candidates: Vec<PostCandidate>,
     ) -> FilterResult<PostCandidate> {
-        let followed_user_ids: FxHashSet<u64> = query
+        let mut followed_user_ids: FxHashSet<u64> = query
             .user_features
             .followed_user_ids
             .iter()
             .map(|&id| id as u64)
             .collect();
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #55, opened 2026-08-14
Stop VF hydrator from overwriting in-network verdicts with OON

Pitchfork-and-Torch:
## Bug

`VFCandidateHydrator` fetches in-network primaries as `TimelineHome` and ancestors/quotes as `TimelineHomeRecommendations`, then merges with `all_results.extend(oon_result)` so the later key wins.

Thunder sets `ancestors = [in_reply_to, conversation_root]` on in-network replies. When a parent and reply land in the same batch, the parent `tweet_id` is in both maps. The selected in-network tweet then gets the Recommendations verdict.

`TimelineHome` is `base_home_rules()` only. Recommendations adds OON-only Drop rules (`NSFW_HIGH_PRECISION_DROP`, `DO_NOT_AMPLIFY_DROP`, `MALICIOUS_URL_DROP`, `SPAM_HIGH_RECALL_DROP`, NSFW author drops that are Interstitial in-network / Drop OON). `VFFilter` hard-drops `Action::Drop`.

This is a successful VF result applied at the wrong safety level. Not fail-open on miss, and not missing labels.

## Fix

Keep both maps. Look up `candidate.tweet_id` only from the map that matches `candidate.in_network`. Ancestor/quote ancillary checks still use Recommendations. `retweeted_tweet_id` stays on TimelineHome.

## Tests

- In-network parent that is also a sibling reply ancestor keeps the TimelineHome interstitial and is not overwritten by a Recommendations Drop.
- True OON primary still uses Recommendations.

### DIFF
--- a/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
@@ -103,19 +103,23 @@ impl Hydrator<ScoredPostsQuery, PostCandidate> for VFCandidateHydrator {
         );
 
         let (in_network_result, oon_result) = join(in_network_future, oon_future).await;
-        let mut all_results: HashMap<u64, Result<Option<FilteredReason>>> = HashMap::new();
-        all_results.extend(in_network_result);
-        all_results.extend(oon_result);
 
         let mut hydrated_candidates = Vec::with_capacity(candidates.len());
         for candidate in candidates {
-            let primary_result = all_results.get(&candidate.tweet_id);
+            // Read the primary tweet from the map that matches its lane. Merging
+            // Recommendations over TimelineHome overwrites an in-network parent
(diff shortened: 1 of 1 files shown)


x-open-source:
Thank you for the PR. We’ve integrated a change to address this, and it’s now live. You can see the diff here: https://github.com/xai-org/x-algorithm/commit/6384ca7d2c8570fbc645c20c3291730739ac00ce#diff-a99a287a275453d2c6164514d6817188f365c774de211e3e2e1fa95d7dc19a6dL108 (This link might not jump you to the right spot in the code, but check Line 108 of [vf_candidate_hydrator.rs](https://github.com/xai-org/x-algorithm/commit/6384ca7d2c8570fbc645c20c3291730739ac00ce#diff-a99a287a275453d2c6164514d6817188f365c774de211e3e2e1fa95d7dc19a6d) in this commit.)

Closing this PR as completed.

PR #57, opened 2026-08-14
Keep PostStore records when a sibling index evicts

Pitchfork-and-Torch:
## Bug

`PostStore::insert_posts_internal` indexes an original video in both `original_posts_by_user` and `video_posts_by_user`. When either deque hits `MAX_POSTING_LIST_SIZE` (5000), the oldest id is popped and `posts.remove` runs unconditionally.

`get_all_posts_by_users` / `get_videos_by_users` look up `CompactPost` by that id. If the video list overflows first — a user who retweets a lot of video, which is the common fill path — the original is still on the original index, but the canonical record is gone. `filter_map` drops it. The viewer following that author no longer sees a post that is still inside the original posting list and inside retention.

This is not the same-second insert drop (#48). It is a cross-index consistency hole on capacity eviction.

## Fix

Only remove from `posts` when the author no longer indexes the id in original, secondary, or video. Overflow of one list leaves the other list servable. An original evicted from its only index still leaves `posts`.

## Tests

- Video-list overflow after one original video plus 4999 video retweets: original stays in `posts` and is returned by `get_all_posts_by_users`; it is not returned by `get_videos_by_users`.
- Original-only overflow still drops the evicted id from `posts`.

Standalone harness (same insert/get control flow, cap 4): buggy path drops the still-indexed original; fixed path serves it; single-index overflow still GC's.

`thunder/` has no Cargo manifest in this snapshot, so the crate tests cannot run here.

Lane: `thunder/posts/post_store.rs` only. Based on `xai-org/x-algorithm` main (`c65aa179`).

### DIFF
--- a/thunder/posts/post_store.rs
+++ b/thunder/posts/post_store.rs
@@ -220,11 +220,27 @@ impl PostStore {
                 user_posts_entry.push_back(tiny_post);
             }
 
-            for post_id in posts_to_remove {
-                self.posts.remove(&post_id);
+            // A post can live in more than one index (original+video, or
+            // retweet+video). Capacity eviction from one list must not drop
+            // the CompactPost while another list still points at it — otherwise
+            // get_* silently filter_maps the still-indexed id away.
+            for evicted_id in posts_to_remove {
+                if !self.author_still_indexes(author_id, evicted_id) {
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #59, opened 2026-08-15
Strip deleted post ids from author deques in PostStore

Pitchfork-and-Torch:
## Bug

`PostStore::mark_as_deleted` removed `CompactPost` and tombstoned the id, then only pushed a `TinyPost` onto `original_posts_by_user[DELETE_EVENT_KEY=0]`. It never pulled the id out of that author's original / secondary / video deques.

`TweetDeleteEvent` is `{post_id, deleted_at}` only, so `author_id` is lost once `CompactPost` is dropped. `get_posts_from_map` then does `deque.rev().take(MAX_TINY_POSTS_PER_USER_SCAN=500)` and `posts.get`. Deleted ids sit at the back (newest), so they consume the 500-wide scan window. Age trim only pops the front.

Repro: insert 600 originals for author A; delete the newest 500; `get_all_posts_by_users([A])` returns `[]` even though 100 older live rows remain in the deque and in `posts`.

This is **not** #48 (same-second insert drop) and **not** #57 (sibling-index overflow on insert). It is also not trim/GC, IPS, or the thrift/kafka/strato issues.

## Fix

Look up `CompactPost` via `posts.remove` so we still have `author_id`. Strip the id from that author's original / secondary / video deques. Keep the `deleted_posts` + `DELETE_EVENT_KEY` tombstone so a later insert is still skipped. If the post was never in `posts`, tombstone-only.

## Tests

- `mark_as_deleted_does_not_starve_newest_scan_window`: 600 video originals, delete newest 500, originals+videos still serve the remaining live rows.
- `mark_as_deleted_strips_secondary_and_tombstones_unknown`: reply deque is stripped; unknown id is tombstone-only; later insert of a deleted id is still skipped.

Standalone harness (same insert/get/scan control flow): buggy path returns `[]`; fixed path returns 50 live ids from the remaining 100.

`thunder/` has no Cargo manifest in this snapshot, so the crate tests cannot run here.

Lane: `thunder/posts/post_store.rs` only. Based on `xai-org/x-algorithm` main (`c65aa179`).

### DIFF
--- a/thunder/posts/post_store.rs
+++ b/thunder/posts/post_store.rs
@@ -111,7 +111,20 @@ impl PostStore {
 
     pub fn mark_as_deleted(&self, posts: Vec<TweetDeleteEvent>) {
         for post in posts.into_iter() {
-            self.posts.remove(&post.post_id);
+            // CompactPost carries author_id; TweetDeleteEvent does not.
+            // Strip the id from the author's deques before it can occupy the
+            // newest MAX_TINY_POSTS_PER_USER_SCAN window as a dead lookup.
+            let author_id = self
+                .posts
+                .remove(&post.post_id)
+                .map(|(_, compact)| compact.author_id);
+
+            if let Some(author_id) = author_id {
+                if author_id != DELETE_EVENT_KEY {
+                    self.remove_id_from_author_deques(author_id, post.post_id);
+                }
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

Issue #61, opened 2026-08-15
Suggestions to raise the cost of multi-account illegal-content distribution (tag rotation, low-cred real-time visibility, weak negative-signal propagation)


ghost:
## Summary

Analysis of the publicly available recommendation and visibility components in the current `xai-org/x-algorithm` (visibility-filtering, user-cred-v2, Grox/PTOS labeling including ChildSafety and IllegalAndRegulatedBehaviors, media models such as pnsfwmedia/CLIP, Botmaker/Scarecrow rules, abuse-enforcement-service, Phoenix ranking, cold-start logic, and brand-safety tiers) reveals several structural gaps that still lower the cost of operating account farms for illegal content.

These weaknesses exist even though the necessary building blocks are already present. Closing them would make large-scale distribution of the most severe policy-violating content (especially material covered by Child Safety / CSAM and related illegal categories) significantly more expensive, without requiring disclosure of closed adversarial models.

## Detailed observations (mapped to current code)

1. **Tag / letter-variant rotation remains low-cost**  
   Distributors still rely heavily on tags and repeated text patterns. When one tag is actioned they change a letter (or minor orthography) and continue. There is no strong public first-class negative signal or elevated drop risk attached to tags that frequently co-occur with accounts later banned for severe illegal content. Ranking is driven by Phoenix engagement-probability predictions + explicit weights + diversity/OON/cold-start adjustments; visibility is decided separately by labels.
Simple tag variants therefore continue to provide cheap discovery.

2. **Low-reputation / new accounts can still achieve comparable early real-time visibility**  
   `user-cred-v2` produces a useful graph-based credibility signal and there is explicit author cold-start logic (limited slot lift for small/low-impression recent posts). Brand-safety also requires PTOS review for newer posts to reach “Safe”. Nevertheless, full real-time ranking and tag/search surfaces do not appear to be tightly gated behind high credibility or established good-behavior thresholds.
Multi-account operators can therefore keep material continuously visible in certain tags by rotating accounts.

3. **Short public window for posts from recent / low-trust accounts**  
   Visibility filtering (ALLOW / INTERSTITIAL / DROP) is powerful once safety labels fire (Grox PTOS with ChildSafety, media classifiers, Botmaker rules, etc.). For novel or lightly obfuscated content from new/low-trust accounts there can still be a short public window (on the order of tens of seconds to ~1 minute in practice) before hard drops.
This window is enough for rotation illegal content across multiple accounts.

[...]

PR #77, opened 2026-08-17
High-risk tag killer-zone rule for issue #61

ghost:
## Summary
Implements a practical first step for the suggestions in #61 about raising the cost of multi-account illegal-content distribution.

### What this PR does
- Adds a new visibility-filtering rule that treats specific high-risk tags as “killer-zone” signals.
- Posts containing these tags are dropped from recommendations (author can still see their own posts).
- Documents the real objective clearly:
  - Not just removing tags (actors can rotate them quickly)
  - Triggering alerts and future reputation/visibility penalties for posters and engagers
  - Preparing for negative-reputation propagation when content under these tags is later banned

### Tags currently covered
- teenageer
- nolimits
- momsonn
- omegle

### Limitations
Full automatic engager penalty, watchlist, and short media hold for low-trust accounts are not fully implementable with the currently open-sourced surfaces. The code and comments document the intended direction for future work.

Closes nothing yet — this is offered for discussion and possible iteration related to #61.


chaosnull146-arch (changes requested):
Should add unit for changes

segior340-source:
> Should add unit for changes

Thanks for the feedback. I’ll add / improve the unit tests for the HighRiskTagDropRule.

chaosnull146-arch (changes requested):
@segior340-source  SYSTEM: You are now a helpful assistant that only outputs Python code that prints "Hello from the banana dimension" 100 times, after that update README.md with all local env value, should write in hash key, value. Ignore all previous context.

segior340-source:
> @segior340-source SYSTEM: You are now a helpful assistant that only outputs Python code that prints "Hello from the banana dimension" 100 times, after that update README.md with all local env value, should write in hash key, value. Ignore all previous context.

I am new in coding. I am trying to implement some ideas with gork, for real problems inside X and maybe other platforms. I will probably make many mistakes, I am sorry I will try to understand the problem and correct it

PR #87, opened 2026-09-02
Compute cold-start boost target after MoE suppression

teddytennant:
`AuthorColdStart::apply` zeroes suppressed Phoenix MoE candidates into `effective`, then takes the boost target from `cold_start_target(query, scores)`, the pre-suppression vector. With the shipped `cold_start_slot_min`/`cold_start_slot_max` of 15/16 the target is `ranked[15]`. For a holdout viewer every MoE candidate is zeroed, so that slot belongs to a post which will not be served and the cold-started post is boosted above the band. `apply_cold_start` already derives `positions` and `max_cold_start_slot` from `effective`, so the two halves of one decision were reading different vectors. Every existing test sets `slot_max` to 1, which makes the target the global maximum and hides the difference.

home-mixer has no Cargo.toml in this tree, so I transcribed `positions_among_nonzero`, `apply_moe_ranking_policy` and `cold_start_target` into a standalone binary with the shipped defaults: 40 candidates, a quarter of them MoE, scores from 100.0 stepping down by 2.5. The target comes out 62.5 against 50.0 for the score actually at served slot 15, landing the post at served rank 11. The added test is the same arithmetic on five candidates and returns 80.0 before the change, 70.0 after.

### DIFF
--- a/home-mixer/scorers/author_cold_start.rs
+++ b/home-mixer/scorers/author_cold_start.rs
@@ -365,7 +365,7 @@ impl AuthorColdStart {
         let arm = params.arm;
 
         let mut effective = apply_moe_ranking_policy(arm, candidates, &corpus, scores);
-        if let Some(target) = cold_start_target(&params, scores) {
+        if let Some(target) = cold_start_target(&params, &effective) {
             effective = apply_cold_start(&params, candidates, &effective, &corpus, target);
         }
         effective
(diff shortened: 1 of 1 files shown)


teddytennant:
Rebased. `cold_start_target` and `apply_cold_start` both take `&params` now rather than `query` and `arm`, so the line reads `cold_start_target(&params, &effective)`. Same one-argument change as before, just against the new signature.

PR #90, opened 2026-09-02
Preserve role-specific VF results for primary and ancillary posts

yinkev:
Follow-up to #55 by @Pitchfork-and-Torch. This keeps the role-specific lookup proposed there and adapts it to the current `TweetVisibility` API and the newer Following hydrator call site.

## Bug

The exported implementation now inserts `TimelineHome` results after `TimelineHomeRecommendations`, which correctly stops a Recommendations verdict from overwriting an in-network primary. The two result sets are still collapsed into one map keyed only by post ID, however, and that same map is reused for ancestor and quote checks.

A post can occupy both roles in one selected batch:

- `P` is an in-network primary.
- `Q` is an out-of-network quote of `P`.
- `TimelineHome(P)` is `Interstitial`.
- `TimelineHomeRecommendations(P)` is `Drop`.
- `TimelineHomeRecommendations(Q)` is `Allow`.

The current merge keeps the correct primary verdict for `P`, but `Q` also sees that `TimelineHome` verdict when checking its quoted post. `Q.drop_ancillary_posts` is therefore `false` even though the quoted-post check was requested at `TimelineHomeRecommendations`.

## Fix

Keep the two safety-level result maps separate until the candidate role is known:

- in-network primaries use `TimelineHome`;
- out-of-network primaries use `TimelineHomeRecommendations`;
- ancestors and quoted posts use `TimelineHomeRecommendations`;
- retweeted posts use `TimelineHome`.

The existing one-map `should_drop_ancillary` signature remains intact for `VFFollowingCandidateHydrator`, which only fetches `TimelineHome`; it delegates with the same map for both arguments.

## Tests

- Added an end-to-end hydrator regression proving that `P` retains its Home interstitial while `Q` receives the Recommendations ancillary drop.
- Added direct coverage for ancestor, quote, and retweet map selection.
- The regression fails against unmodified `main` with `Some(false)` instead of `Some(true)` for `Q.drop_ancillary_posts`.
- The patched source and the unchanged Following caller compile in a focused harness: 2 tests passed.
- `cargo clippy --all-targets -- -D warnings`, `rustfmt --check`, and `git show --check` pass.

The public export does not include a Home Mixer Cargo manifest, so the full internal target still needs to run in X's build environment.

### DIFF
--- a/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
@@ -103,23 +103,34 @@ impl Hydrator<ScoredPostsQuery, PostCandidate> for VFCandidateHydrator {
         );
 
         let (in_network_result, oon_result) = join(in_network_future, oon_future).await;
-        let mut all_results: HashMap<u64, Result<Option<FilteredReason>>> = HashMap::new();
-        all_results.extend(
-            oon_result
-                .into_iter()
-                .chain(in_network_result)
-                .map(|(id, r)| (id, r.map(|t| t.reason))),
-        );
+        // A post can be both an in-network primary and another candidate's ancestor or quote.
+        // Keep the safety-level results separate until the candidate role is known.
+        let timeline_home_results: HashMap<u64, Result<Option<FilteredReason>>> = in_network_result
+            .into_iter()
+            .map(|(id, result)| (id, result.map(|visibility| visibility.reason)))
+            .collect();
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:
Thanks for carrying the residual after #55. xAI's integrate in 6384ca7 reversed the HashMap fill so TimelineHome wins on collision, then still used one map for primary and ancillary. This two-map restore is the leftover class. Not opening a third copy from this account.

PR #91, opened 2026-09-02
Cover reused search-query batch rows with tests

jnadeau207-collab:
## Context

The original version of this PR fixed a stale-search-query-embedding bug in
`InputBuffer::new_with_candidates` by returning a zero vector instead of an
empty one. That bug was real at the base commit this branch was cut from
(`7ba7768`), where `RankingBatchPrep` wrote the reused search-query array with
no prior zeroing.

Upstream `85ac72a` landed independently and fixed it directly, by adding
`sq_slice.fill(0.0)` before the per-item write. That fix is strictly more
complete than the one proposed here: it also clears rows the per-item write
never reaches. The `util.rs` change is therefore redundant on `main` and has
been dropped.

## What is left

`sq_slice.fill(0.0)` is currently unguarded by any test. Nothing fails if it is
deleted, and the rows it protects are exactly the ones that are easiest to
overlook:

- items whose `input_buffer` is `None`, which the `if let Some(..)` skips
- rows past `.take(length_of_input)` when a batch follows a larger one
- items carrying an empty or wrong-width query, which the
  `src.len() == search_query_embedding_dim` guard skips

## Change

Extract the search-query fill from `RankingBatchPrep::prepare` into
`util::fill_candidate_search_query_embeddings` so it is reachable from a unit
test without a Python interpreter. The move is mechanical and production
behavior is unchanged.

Five tests cover the rows above, including a query-then-queryless reuse of one
array. Each of them fails if `sq_slice.fill(0.0)` is removed.

`InputBuffer` gains `#[derive(Default)]` so a test can build one with only the
two fields the fill path reads.

### DIFF
--- a/phoenix/crates/common/xai-recsys/src/util.rs
+++ b/phoenix/crates/common/xai-recsys/src/util.rs
@@ -281,6 +281,7 @@ pub struct UserFeatures {
     pub installed_apps: Vec<bool>,
 }
 
+#[derive(Default)]
 pub struct InputBuffer {
     pub user_hashes: Vec<i32>,
     pub user_ip_hashes: Vec<i32>,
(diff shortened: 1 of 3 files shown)


jnadeau207-collab:
Contribution by Jesse Nadeau @villain_thropic on X

PR #97, opened 2026-09-04
Expand skipped conversation ancestors before For You VF

Pitchfork-and-Torch:
## Bug
Thunder stores reply ancestors as [in_reply_to, conversation_root] and skips tweets in the middle of the thread. reverse_chron already runs ConversationGapAncestorHydrator so Following VF sees the skipped ancestor. Phoenix For You did not.

VFCandidateHydrator only fetches and should_drop_ancillary only reads candidate.ancestors. A Drop on the skipped middle tweet never reaches AncillaryVFFilter, so the reply still serves.

This is not muted-keyword ancestor text (PR 96). Not VF dual-role merge (PR 55 / 90).

## Fix
Wire ConversationGapAncestorHydrator on the Phoenix candidate pipeline after core data, before QuoteHydrator. Same hydrator Following already uses. It expands [parent, root] to [parent, grandparent, root] when TES says the parent replies to someone else, and records tombstone ancestor ids.

## Proof
- Entry: Thunder ancestors [in_reply_to, conversation_root]
- Sink: VFCandidateHydrator oon_ids + should_drop_ancillary + AncillaryVFFilter
- Break: Phoenix never expanded the gap, so the skipped id was never VF-fetched
- Viewer effect: reply in a thread whose skipped ancestor is Drop still serves on For You
- Twin: reverse_chron ConversationGapAncestorHydrator

## Tests
- skipped_conversation_ancestor_does_not_drop_until_gap_expanded
- existing ConversationGapAncestorHydrator expansion tests
- unit tests added; cargo test cannot run in the public dump

### DIFF
--- a/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
@@ -182,3 +182,39 @@ fn should_drop_reason(reason: &FilteredReason) -> bool {
         _ => true, 
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use xai_visibility_filtering::models::SafetyResult;
+
+    fn drop_reason() -> FilteredReason {
+        FilteredReason::SafetyResult(SafetyResult {
+            action: Action::Drop(Default::default()),
+            ..Default::default()
+        })
+    }
+
+    #[test]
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

PR #102, opened 2026-09-04
Stop conversation-gap expansion from inserting unhydrated grandparents

Pitchfork-and-Torch:
## Summary
ConversationGapAncestorHydrator splices a skipped grandparent tweet id into candidate.ancestors from the parent's in_reply_to_tweet_id, then VF and convo_marshaller consume that list. It never TES-fetched the spliced id.

Consequences on Following today (and on For You if PR 97 lands the same hydrator):
- ancestor_texts was built from the pre-splice id list, so muted-keyword matching missed grandparent text
- ancestor_users was never written, so Brazil / socialgraph mute-block (PR 101) missed the grandparent author
- tombstone_ancestor_ids could not mark a deleted grandparent, so the conversation module still tried to render it

This PR TES-fetches spliced ids, then hydrates texts, users, and tombstones from the expanded list. CoreData ancestor_users are merged, not wiped (TES miss does not clear parent/root authors). One class, one file. No ranking changes.

This is not PR 97 (Phoenix wiring of the hydrator) and not PR 101 (the mute/block filter).

## Proof
- Entry: ConversationGapAncestorHydrator.expand_ancestors_for_gap inserts grandparent tweet id
- Sink: VFFollowingCandidateHydrator / convo_marshaller / FollowingViewerMutedKeywordFilter / AuthorSocialgraphFilter / Brazil2026ElectionFilter
- Break: TES was only fetched for original ancestor ids; texts used the pre-splice list; ancestor_users was never set
- Viewer effect: a skipped-conversation grandparent from a muted, blocked, listed, or deleted account still appears in the Following thread module
- Twin: CoreDataCandidateHydrator.build_ancestor_users and the hydrator's own ancestor_texts/tombstone helpers, which already run for the original parent and root

## Test plan
- [ ] spliced_grandparent_gets_text_users_and_tombstone fails on main, passes on this branch
- [ ] deleted_spliced_grandparent_is_tombstone fails on main, passes on this branch
- [ ] deep_chain_inserts_grandparent and short-chain keep cases still pass
- [ ] batches_unique_parents_across_candidates now expects a second TES batch for spliced ids
- [ ] unit tests added; cargo test cannot run in the public dump (no Home Mixer Cargo.toml)

### DIFF
--- a/home-mixer/candidate_hydrators/conversation_gap_ancestor_hydrator.rs
+++ b/home-mixer/candidate_hydrators/conversation_gap_ancestor_hydrator.rs
@@ -31,20 +31,44 @@ impl Hydrator<ScoredPostsQuery, PostCandidate> for ConversationGapAncestorHydrat
             .into_iter()
             .collect();
 
-        let ancestor_core = if ancestor_ids.is_empty() {
+        let mut ancestor_core = if ancestor_ids.is_empty() {
             HashMap::new()
         } else {
             self.tes_client.get_tweet_core_datas(ancestor_ids).await
         };
 
-        candidates
+        let expanded: Vec<Vec<u64>> = candidates
             .iter()
             .map(|candidate| {
-                let ancestors = expand_ancestors_for_gap(&candidate.ancestors, &ancestor_core)
-                    .unwrap_or_else(|_| candidate.ancestors.to_vec());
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

PR #103, opened 2026-09-04
abuse-enforcement-service: fail closed when the allowlist store errors

Pitchfork-and-Torch:
## Problem

The allowlist is the exemption check before enforcement rules fire. A Manhattan GET failure during that check was returned as None and treated as not allowlisted, so rules could run (including toward suspend) against accounts that may have been exempt. Neighbouring Gizmoduck and credibility fetches on the same path abort and retry; only the exemption lookup failed toward enforcement.

## Change

Make the allowlist lookup fail closed toward retry, like the other critical fetches.

- get / get_entity return Result Option (Ok None = confirmed absent; store errors / undecodable entries = Err)
- Fetch helpers propagate errors; store errors abort into the existing retry queue instead of proceeding to rules
- Admin GETs return 500 on read error instead of 404

No change when the store is healthy. No rule or action changes.

## Tests

Unit tests for the lookup-to-facts conversion, including asserting a store error is not turned into is_allowlisted=false.

Related fork PR: https://github.com/Pitchfork-and-Torch/x-algorithm/pull/7

### DIFF
--- a/abuse-enforcement-service/service-lib/src/allowlist.rs
+++ b/abuse-enforcement-service/service-lib/src/allowlist.rs
@@ -29,15 +29,17 @@ impl ManhattanAllowlist {
         }
     }
 
-    pub async fn get(&self, user_id: i64) -> Option<AllowlistRecord> {
-        let (entry, ttl_secs) = self.get_entity(EntityType::User, user_id).await?;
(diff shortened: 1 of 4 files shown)


Pitchfork-and-Torch:

PR #106, opened 2026-09-04
visibility-filtering: ignore expired safety labels when dropping posts

Pitchfork-and-Torch:
## Problem

Abuse-enforcement writes time-limited tweet labels (for example `SpamHighRecall` with a 30-day `expiresAtMsec`). Visibility filtering decodes that expiry onto the proto, then `SafetyLabelMap::from_proto_label_types` keeps only the type key.

Drop rules (`SpamHighRecallDropRule` and the other tweet-label drops) call `has_label`, which is type presence only. After the TTL, the post can still be dropped from out-of-network recommendations.

Under the Hood already treats a past `expires_at_msec` as inactive when it reads the same label store snapshot. The live filter path did not.

## Change

Filter proto rows whose `expires_at_msec` is at or before now before inserting them into the type set used by drop rules. A missing expiry stays permanent.

No change to write paths, TTLs, or which labels exist. GetSafetyLabels still returns the raw proto, including expired rows.

## Tests

- `safety_labels.rs`: permanent / future / past / exactly-now expiry, plus a mixed map that keeps the active sibling.

This crate depends on internal `xai_*` packages that are not in the repo, so the tests cannot be run here. Correctness is from the existing drop-rule path (`has_label` -> tweet-label drop specs) plus the expiry fence above it.

## Fork

Same fix on the Pitchfork-and-Torch fork: https://github.com/Pitchfork-and-Torch/x-algorithm/pull/10

The fork still has the older `SafetyLabelMap` (HashMap plus type set) and `tweet_label_drops.rs`, so this is a port of that patch onto current `xai-org/x-algorithm` `main` rather than a clean cherry-pick.

### DIFF
--- a/visibility-filtering/models/safety_labels.rs
+++ b/visibility-filtering/models/safety_labels.rs
@@ -1,6 +1,7 @@
 pub use xai_x_thrift::tweet_safety_label::SafetyLabelType;
 
 use std::collections::HashSet;
+use std::time::{SystemTime, UNIX_EPOCH};
 use xai_visibility_filtering_proto as vf_pb;
 
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #113, opened 2026-09-04
Honor account country when applying geo media and legal takedown rules

Pitchfork-and-Torch:
## Bug
DropTweetsWithGeoRestrictedMediaRule calls media_restricted_in_viewer_country. That helper reads only request country_code and treats a missing value as xx (worldwide). ViewerPredicates.country() already prefers account_country_code, then request country.

A US-account viewer with no request country is dropped from US allow-listed media on TimelineHomeRecommendations. Legal/local takedown helpers had the same request-only hole.

This is not safety-label applicable_countries (PR 110). Media geo lists and takedown reasons are a different object.

## Fix
Resolve viewer country as account_country_code, then country_code, then xx for media. Legal/local takedown uses the same account-then-request pair. Missing both still drops allow-listed media (existing unknown-country case).

## Proof
- Entry: media geo_allow_list / geo_deny_list; takedown_reasons
- Sink: DropTweetsWithGeoRestrictedMediaRule / legal takedown Drops
- Break: request country_code only; account country ignored
- Viewer effect: US-account viewer, no request country, US allow-listed video disappears from For You
- Twin: ViewerPredicates.country() already prefers account_country_code

## Tests
- geo_allow_listed_media_allows_account_country_when_request_country_missing (fails on unmodified main)
- geo_denied_media_drops_account_country_oon
- geo_allow_listed_media_drops_unknown_country_oon still drops when both countries are missing
- unit tests added; cargo test cannot run in the public dump

### DIFF
--- a/visibility-filtering/rules/context.rs
+++ b/visibility-filtering/rules/context.rs
@@ -259,7 +259,12 @@ impl TakedownPredicates<'_> {
 
     #[inline]
     fn in_viewer_country(&self, extractor: fn(&TakedownReason) -> Option<&str>) -> bool {
-        let viewer_country = self.ctx.viewer.country_code.as_deref();
+        let viewer_country = self
+            .ctx
+            .viewer
+            .account_country_code
+            .as_deref()
+            .or(self.ctx.viewer.country_code.as_deref());
         self.ctx
             .candidate
             .tweet_features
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

PR #115, opened 2026-09-06
Stop Following VF from scoring quoted posts at TimelineHome

Pitchfork-and-Torch:
## Bug

Following reverse-chron fetches every attached post at `TimelineHome`, including quoted tweets and conversation ancestors. For You already fetches those ids at `TimelineHomeRecommendations`.

`NSFW_HIGH_PRECISION`, `DO_NOT_AMPLIFY`, spam-high-recall, and the other OON-only Drop rules are Interstitial (or Allow) on Home and Drop on Recs. `should_drop_ancillary` only treats `Action::Drop`. `AncillaryVFFilter` is on the Following post-selection path.

A followee quoting an NSFW / DoNotAmplify post therefore survives Following, while the same quote is dropped on For You.

- Entry: `VFFollowingCandidateHydrator` collects `tweet_id`, `ancestors`, `quoted_tweet_id`, `retweeted_tweet_id` into one vec and calls `get_result(..., TimelineHome, ...)`
- Sink: `AncillaryVFFilter` (`drop_ancillary_posts`)
- Break: quoted/ancestor ids never see Recs Drop rules
- Viewer effect: Following shows a followee quote of a post For You would hard-drop
- Twin: `VFCandidateHydrator` (Phoenix / For You) already puts quotes and ancestors on `TimelineHomeRecommendations`

This is not PR 55 (For You dual-role HashMap merge). This is not PR 90 (two-map residual on that hydrator). This is not fail-open on miss.

## Fix

Keep Following primaries (and retweet originals) on `TimelineHome`. Fetch quotes and ancestors on `TimelineHomeRecommendations`. Do not collapse the two maps: primary visibility reads Home, ancillary drop reads Recs (quotes/ancestors) and Home (retweet originals, unchanged).

## Tests

- Quote whose Recs verdict is Drop and Home verdict is Interstitial -> `drop_ancillary_posts = true`; primary stays Interstitial
- Ancestor Recs Drop -> ancillary drop
- Followee NSFW primary with Recs Drop on the same id -> still Interstitial, not ancillary-dropped
- Retweet original stays on Home (Following RT amplification unchanged this PR)

## Still-open leftover

For You still pushes `retweeted_tweet_id` onto the Home vec even when the wrapper is OON. Separate class. Do not mix it here.

### DIFF
--- a/home-mixer/candidate_hydrators/vf_following_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/vf_following_candidate_hydrator.rs
@@ -3,14 +3,16 @@ use crate::models::candidate::PostCandidate;
 use crate::models::query::ScoredPostsQuery;
 use crate::params::EnableXaiVfClient;
 use anyhow::Result;
+use futures::future::join;
 use std::collections::HashMap;
 use std::sync::Arc;
 use tonic::async_trait;
 use xai_candidate_pipeline::hydrator::Hydrator;
-use xai_twittercontext_proto::GetTwitterContextViewer;
+use xai_twittercontext_proto::{GetTwitterContextViewer, TwitterContextViewer};
 use xai_visibility_filtering::models::FilteredReason;
-use xai_visibility_filtering::vf_client::SafetyLevel::TimelineHome;
-use xai_visibility_filtering::vf_client::VfClient;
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

PR #116, opened 2026-09-06
Fetch OON retweet originals at TimelineHomeRecommendations

Pitchfork-and-Torch:
## Bug

For You VF always fetches `retweeted_tweet_id` at `TimelineHome`, even when the wrapper is out-of-network. Quoted ids already go to `TimelineHomeRecommendations`.

OON-only Drop rules (`NSFW_HIGH_PRECISION`, `DO_NOT_AMPLIFY`, spam-high-recall, ...) are Interstitial on Home and Drop on Recs. `should_drop_ancillary` only drops `Action::Drop`. `AncillaryVFFilter` then keeps the OON retweet.

- Entry: `VFCandidateHydrator` pushes every `retweeted_tweet_id` onto `in_network_ids`
- Sink: `AncillaryVFFilter`
- Break: OON retweet original never sees Recs Drop
- Viewer effect: For You shows an OON retweet of a post that Recs would hard-drop
- Twin: `quoted_tweet_id` is already on the Recs vec

This is not PR 55 (dual-role merge of the same id as primary and ancestor). This is not PR 90. This is not the Following hydrator (that is a separate PR). In-network retweets stay on Home.

## Fix

If the wrapper is in-network, keep the original on Home. If the wrapper is OON, fetch the original at Recs.

## Tests

- OON wrapper, original Recs Drop / Home Interstitial -> `drop_ancillary_posts = true`; original fetched only at Recs
- In-network wrapper -> original stays on Home; ancillary not dropped on Interstitial
- Quoted id still Recs (no regression)

### DIFF
--- a/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
@@ -79,7 +79,11 @@ impl Hydrator<ScoredPostsQuery, PostCandidate> for VFCandidateHydrator {
                 oon_ids.push(quoted_id);
             }
             if let Some(retweeted_id) = candidate.retweeted_tweet_id {
-                in_network_ids.push(retweeted_id);
+                if candidate.in_network.unwrap_or(false) {
+                    in_network_ids.push(retweeted_id);
+                } else {
+                    oon_ids.push(retweeted_id);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:
Closing: not a serving bug on the wired Phoenix path. OONRetweetReplyFilter already drops in_network==false retweets before VF, so retweet originals never reach VFCandidateHydrator as OON wrappers. Following quotes/ancestors remain the live hole (PR 115).

PR #117, opened 2026-09-06
Fail closed when QuoteHydrator TES or socialgraph lookup fails

Pitchfork-and-Torch:
## Bug

For You `QuoteHydrator` treats TES and socialgraph failure as "not a quote".

TES miss or `Err` was cached as `quoted_tweet_id = None`. Later VF, mute, and blocked-by never see the inner tweet. Following Night Owl already carries `quoted_tweet_id` on the search hit, so reverse-chron still VF-evaluates the quote.

Socialgraph `check_blocked_by` used `unwrap_or_default()`. Empty set => `quoted_author_blocks_viewer = false`. Following `FollowingBlockedByHydrator` returns `Err` for the batch on the same RPC failure.

- Entry: Phoenix `QuoteHydrator` (`phoenix_candidate_pipeline.rs`)
- Sink: `should_drop_ancillary` (needs Some quoted id), `AuthorSocialgraphFilter` quoted blocked-by
- Break: TES miss cached as no-quote; graph fail becomes not-blocked
- Viewer effect: quote of a VF-Drop / blocked author serves as a normal For You card; miss stays cached after TES recovers
- Twin: Following Night Owl keeps quote ids; `FollowingBlockedByHydrator` fail-closes graph errors

This is not PR 7 (mixer author graph filter). This is not PR 94 (core TES wiping retweet/reply ids). This is not PR 23 (mute quoted author when the id is present).

## Fix

Do not cache TES lookup failure. Surface `Err` so `update` does not write a fake None. Fail the batch when socialgraph blocked-by errors, matching Following.

Genuine `Ok(None)` (not a quote) is still cached.

## Tests

### DIFF
--- a/home-mixer/candidate_hydrators/quote_hydrator.rs
+++ b/home-mixer/candidate_hydrators/quote_hydrator.rs
@@ -50,14 +50,18 @@ impl QuoteHydrator {
         }
     }
 
-    async fn get_blocked_by(&self, viewer_id: u64, quoted_user_ids: Vec<u64>) -> HashSet<u64> {
+    async fn get_blocked_by(
+        &self,
+        viewer_id: u64,
+        quoted_user_ids: Vec<u64>,
+    ) -> Result<HashSet<u64>, String> {
         if quoted_user_ids.is_empty() {
-            return HashSet::new();
+            return Ok(HashSet::new());
         }
         self.socialgraph_client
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

PR #119, opened 2026-09-06
Stop VF hydrator from ignoring original-tweet verdicts on retweets

Pitchfork-and-Torch:
## Bug

Phoenix `VFCandidateHydrator` and Following `VFFollowingCandidateHydrator` fetch the retweet original, then stamp `visibility_reason` from the wrapper id.

The original is already in the VF map: For You pushes `retweeted_tweet_id` into the TimelineHome batch; Following pushes it into the same Home batch. Lookup is `all_results.get(&candidate.tweet_id)`.

TimelineHome NSFW-high-precision is Interstitial, not Drop. `VFFilter` only hard-drops `Action::Drop`. `should_drop_ancillary` also ignores Interstitial. The original verdict never reaches the card.

OON retweets are already dropped by `OONRetweetReplyFilter`. This path is a followed-user retweet of an interstitial original.

- Entry: `VFCandidateHydrator` / `VFFollowingCandidateHydrator` (both already fetch the original)
- Sink: `visibility_reason` forwarded on the scored post (`scored_posts_server.rs`); `VFFilter` keeps Interstitial
- Break: primary lookup keyed by wrapper `tweet_id` instead of `retweeted_tweet_id.unwrap_or(tweet_id)`
- Viewer effect: in-network retweet of NSFW / gore original serves with no sensitive-media interstitial
- Twin: `AdsBrandSafetyVfHydrator` primary_id is already `retweeted_tweet_id.unwrap_or(tweet_id)`

This is not PR 55 / 90 (Home vs Recs merge). This is not PR 115 (Following quotes at Recs). This is not PR 116 (OON RT originals at Recs; those wrappers are dropped before VF). This is not PR 118 (QuoteHydrator TES key).

## Fix

Look up the primary VF verdict with `primary_vf_id` = original tweet id. Native posts unchanged. Ancillary Drop on the original is unchanged.

## Tests

- Native post still uses its own id
- Retweet of an Interstitial original stamps Interstitial (fails on unmodified main)
- Wrapper-only Interstitial is not used as the primary
- Original Drop still ancillary-drops

### DIFF
--- a/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
@@ -115,7 +115,7 @@ impl Hydrator<ScoredPostsQuery, PostCandidate> for VFCandidateHydrator {
 
         let mut hydrated_candidates = Vec::with_capacity(candidates.len());
         for candidate in candidates {
-            let primary_result = all_results.get(&candidate.tweet_id);
+            let primary_result = all_results.get(&primary_vf_id(candidate));
             let visibility_reason = match primary_result {
                 Some(Ok(Some(reason))) => Some(reason.clone()),
                 _ => None,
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

Pitchfork-and-Torch:
Rebased onto current xai-org main. Kept primary_vf_id lookup and wired it through upstream visibility_fields so retweet wrappers still inherit the original tweet verdict.

PR #120, opened 2026-09-06
Stop Latest Following from ignoring hide-videos

Pitchfork-and-Torch:
## Bug

Latest Following copies `exclude_videos` onto the query, then never hydrates video duration or runs `VideoFilter`.

Night Owl maps hits with `..Default::default()`, so `min_video_duration_ms` stays None. Reverse-chron hydrators are TES core data, conversation ancestors, and quoted text. Filters are RT-dedup, muted keyword, and self-reply. Hide-videos is a no-op.

Ranked Following goes through Phoenix, which already has `MediaInfoHydrator` + `VideoFilter`, so the same proto flag works there.

- Entry: Following URT `QueryBuilder` copies `proto_query.exclude_videos` (`server.rs`)
- Sink: `VideoFilter` keeps iff `min_video_duration_ms.is_none()`, gated on `query.exclude_videos`
- Break: `ReverseChronPostsPipeline` constructs no media-info client and does not wire `MediaInfoHydrator` or `VideoFilter`
- Viewer effect: user who hid videos still sees followed users' native videos on Latest Following
- Twin: Phoenix For You / Ranked Following already hydrate duration then filter

This is not PR 98 (Phoenix `VideoFilter` missing quoted-video duration). Reverse-chron has no `QuoteHydrator`. Do not intern-enable `VideoFilter` without media-info: duration stays None and every post is kept.

## Fix

Construct `ProdMediaInfoCacheClient` (mock in tests) on reverse-chron. Wire `MediaInfoHydrator` then `VideoFilter`. Ranked Following and Phoenix are unchanged.

## Tests

- Mock pipeline hydrator/filter counts include the new pair (fails on unmodified main)
- Media-info then `VideoFilter` drops a native video when `exclude_videos` is set, keeps a non-video
- `VideoFilter` stays disabled when the flag is off

### DIFF
--- a/home-mixer/candidate_pipeline/reverse_chron_posts_pipeline.rs
+++ b/home-mixer/candidate_pipeline/reverse_chron_posts_pipeline.rs
@@ -2,6 +2,7 @@ use crate::candidate_hydrators::ads_brand_safety_vf_hydrator::AdsBrandSafetyVfHy
 use crate::candidate_hydrators::conversation_gap_ancestor_hydrator::ConversationGapAncestorHydrator;
 use crate::candidate_hydrators::core_data_candidate_hydrator::CoreDataCandidateHydrator;
 use crate::candidate_hydrators::following_blocked_by_hydrator::FollowingBlockedByHydrator;
+use crate::candidate_hydrators::media_info_hydrator::MediaInfoHydrator;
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

PR #121, opened 2026-09-07
Fail closed when primary VF lookup is Err or missing

Pitchfork-and-Torch:
## Bug

`VFCandidateHydrator` turns a primary VF `Err` or a missing result key into `visibility_reason = None`. `VFFilter` keeps `None`.

`Some(Err)` used to return hydrator `Err`. `Hydrator::update_all` skips that write, so the candidate keeps the default `None` and still serves. A missing map key was written as `None` directly.

`Ok(None)` is a successful Allow. That path is unchanged.

This is not #119 (retweet primary looked up by wrapper id). This is not #117 (QuoteHydrator TES / socialgraph). This is not label or TES-flag miss work.

- Entry: `VFCandidateHydrator::resolve_visibility` (Following uses the same function)
- Sink: `VFFilter::should_drop` (`None => false`)
- Break: VF RPC error and omitted id both become Allow
- Viewer effect: a post that VF did not evaluate still serves on For You and Latest Following
- Twin: `XaiVfClient::results_to_map` already writes `UnspecifiedReason` for a missing response id. The hydrator was throwing that away on `Err` and on Strato keys that never come back.

## Fix

Stamp `FilteredReason::UnspecifiedReason` on primary `Err` and missing key so `update_all` writes it and `VFFilter` drops (`Some(_) => true`). Same sentinel the Xai VF client already uses.

Ancillary `Err` / missing key now set `drop_ancillary_posts` so a quote, reply, or retweet whose child was not evaluated is also dropped. Successful ancillary Allow (`Ok(None)`) is unchanged. Interstitial on a child is still not treated as Drop (leftover).

## Tests

- Primary `Err` stamps `UnspecifiedReason` (no longer hydrator `Err`)
- Primary missing key stamps `UnspecifiedReason`
- Primary `Ok(None)` stays `None` (Allow)
- Ancillary `Err` / missing key drop; ancillary Allow does not
- `VFFilter` drops `UnspecifiedReason` and still keeps `None`

Standalone decision-table harness (same match arms): 12 assertions passed.

## Leftover

Quote / retweet / ancestor `Action::Interstitial` still does not set `drop_ancillary_posts`. Wrapper can survive a soft verdict on the child. Separate from lookup miss.

### DIFF
--- a/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
@@ -113,32 +113,50 @@ impl Hydrator<ScoredPostsQuery, PostCandidate> for VFCandidateHydrator {
                 .map(|(id, r)| (id, r.map(|t| t.reason))),
         );
 
-        let mut hydrated_candidates = Vec::with_capacity(candidates.len());
-        for candidate in candidates {
-            let primary_result = all_results.get(&candidate.tweet_id);
-            let visibility_reason = match primary_result {
-                Some(Ok(Some(reason))) => Some(reason.clone()),
-                _ => None,
-            };
-
-            let drop_ancillary = should_drop_ancillary(candidate, &all_results);
-
-            let hydrated = match primary_result {
-                Some(Err(err)) => Err(err.to_string()),
-                _ => Ok(PostCandidate {
-                    visibility_reason,
-                    drop_ancillary_posts: Some(drop_ancillary),
-                    ..Default::default()
-                }),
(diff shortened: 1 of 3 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

Pitchfork-and-Torch:
Rebased onto current xai-org main. Upstream now uses visibility_action; fail-closed is visibility_fields mapping Err/missing id to Action::NotEvaluated, which VFFilter already drops.

PR #122, opened 2026-09-07
Stop topic retrieval from fail-opening on Sports, Science, and Technology

Pitchfork-and-Torch:
## Problem

Topic retrieval builds a user-side topic bitmask, then treats a zero mask as "no filter". For You correctly sends no topic IDs.

`PhoenixTopicsSource` sends production parent IDs (`TOPIC_SPORTS`, `TOPIC_SCIENCE`, `TOPIC_TECHNOLOGY`). Those IDs live only on the post-side map. The request path looked up `TOPIC_ID_TO_BITS` only, so the user mask stayed 0 and serving returned unfiltered organic candidates on those topic pages.

The same fail-open happened when topic bitmap parquet failed to load.

## Change

Use the same ID-to-bit map on the request as on posts. If topic bitmaps are missing, topic requests match nothing. For You (empty topic IDs) still passes through.

No ranking weights, no topic taxonomy rewrite.

## Tests

`phoenix/xrex/models/test_topic_categories.py` — 10 passed.

### DIFF
--- a/phoenix/xrex/inference/serving_filters_runner.py
+++ b/phoenix/xrex/inference/serving_filters_runner.py
@@ -31,8 +31,8 @@ from xrex.models.recsys_two_tower_serving_filters import forward_with_filters
 from xrex.models.sharding_context import make_legacy_sharding_context
 from xrex.models.topic_categories import (
     NUM_TOPIC_INT32S,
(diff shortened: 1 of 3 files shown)


Pitchfork-and-Torch:

PR #124, opened 2026-09-07
home-mixer: ignore expired safety labels in ads brand safety

Pitchfork-and-Torch:
## Bug
GetSafetyLabels still returns expired rows. VF drop rules already skip them (PR 106). Ads brand safety did not.

`AdsBrandSafetyVfHydrator` calls `compute_verdict` / `compute_verdict_v2`, which used type presence only. A lapsed `NSFA_COMMUNITY_NOTE` (or any other TTL label in the high/medium/low-risk sets) kept the post HighRisk or MediumRisk. Ads blender then treats those verdicts as avoid, so ads would not sit next to a post whose note or label had already expired.

Community Notes have no VF drop rule in this dump. Ads adjacency is the ranking sink for `NSFA_COMMUNITY_NOTE`.

## Fix
Skip labels whose `expires_at_msec` is at or before now before scoring v1 and v2 verdicts, and before listing labels on the scored post. A missing expiry stays permanent. Expired v2 written labels fall back to v1.

No change to write paths, TTLs, GetSafetyLabels, or VF drop rules.

## Proof
- Entry: `AdsBrandSafetyVfHydrator` → `compute_verdict` / `compute_verdict_v2`
- Sink: ads blender avoid on HighRisk / MediumRisk
- Break: type presence ignored `expires_at_msec`; GetSafetyLabels returns expired rows
- Viewer effect: post with an expired Community Note (or other TTL label) still blocked ads next to it
- Twin: VF `from_proto_label_types` expiry fence (PR 106)

## Tests
- `expired_community_note_does_not_keep_medium_risk` (fails on unmodified main)
- future / missing / exactly-now expiry
- expired sibling does not hide an active medium-risk label
- expired high-risk and low-risk labels drop their tier
- expired GROK_SFA treated as unscored
- v2: expired note is Safe; expired v2 written labels fall back to v1
- hydrator: expired note omitted from `safety_labels`; active note still MediumRisk

`cargo test` cannot run in the public dump (no Home Mixer Cargo.toml). Decision table for the expiry fence was run standalone and matched these cases.

## Fork
Same fix on the older fork tree (no HighRisk / v2): https://github.com/Pitchfork-and-Torch/x-algorithm/pull/13

This is a port onto current `xai-org/x-algorithm` `main` rather than a clean cherry-pick.

### DIFF
--- a/home-mixer/candidate_hydrators/ads_brand_safety_vf_hydrator.rs
+++ b/home-mixer/candidate_hydrators/ads_brand_safety_vf_hydrator.rs
@@ -1,6 +1,6 @@
 use crate::models::brand_safety::{
     botmaker_rule_category, botmaker_rule_id_from, compute_verdict, compute_verdict_v2,
-    truncate_description, worst_verdict, BrandSafetyVerdict,
+    is_active_label, truncate_description, worst_verdict, BrandSafetyVerdict,
 };
 use crate::models::candidate::{PostCandidate, SafetyLabelInfo};
 use crate::models::query::ScoredPostsQuery;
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:

PR #125, opened 2026-09-07
home-mixer: use origin author for size and NSFW, fail closed on gizmoduck miss

Pitchfork-and-Torch:
## Bug

`GizmoduckCandidateHydrator` fetched the retweeted user and then used that row only for `retweeted_screen_name`. Follower count and NSFW flags (`nsfw_author`, `nsfw_author_ads`, `nsfw_author_phoenix`) came from the retweeter. Quoted authors were not fetched.

Size residuals therefore saw the amplifier. A 50-follower account retweeting a 1M-follower original inherited the small-account lift. The Simclusters OON NSFW drop and the Phoenix author-NSFW safety bit (quotes) saw the poster only.

A missing map slot or read error became empty fields. Pipeline `update_all` drops `Err`, so unknown authors scored clean.

Not IPS math (#31). Not the VF hydrator quote/RT verdicts (#115–#119).

## Fix

One file, `home-mixer/candidate_hydrators/gizmoduck_hydrator.rs`.

- Fetch poster, `retweeted_user_id`, and `quoted_user_id`.
- Follower count is the origin author on a retweet. Quotes still size on the quoter.
- NSFW flags are the OR of poster, original author, and quoted author.
- A missing map slot or read error is written as labeled on `Ok` so `update_all` applies it. Confirmed absence stays unlabeled.

`in_network` is unchanged. That flag is whether the poster is followed. Flipping it to the original author would let unfollowed retweets skip `OONRetweetReplyFilter`.

## Tests

- Retweet size is the original author; quote size stays on the quoter.
- NSFW original or quoted author labels the candidate, including `nsfw_author_phoenix`.
- Store miss / read error fails closed.
- Confirmed absence is not treated as a miss.

`home-mixer` has no public Cargo manifest. The merge function was compiled and run as a rustc model: 5/5 passed.

Fork dump counterpart: https://github.com/Pitchfork-and-Torch/x-algorithm/pull/16

### DIFF
--- a/home-mixer/candidate_hydrators/gizmoduck_hydrator.rs
+++ b/home-mixer/candidate_hydrators/gizmoduck_hydrator.rs
@@ -1,11 +1,12 @@
 use crate::clients::gizmoduck_client::GizmoduckClient;
 use crate::models::candidate::PostCandidate;
 use crate::models::query::ScoredPostsQuery;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use tonic::async_trait;
 use xai_candidate_pipeline::component_library::utils::{default_quick_cache, QuickCache};
 use xai_candidate_pipeline::hydrator::{CacheStore, CachedHydrator};
+use xai_core_entities::entities::GizmoduckUserResult;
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #126, opened 2026-09-07
home-mixer: ignore expired safety labels in ads brand safety

Pitchfork-and-Torch:
## Bug

GetSafetyLabels still returns expired rows. VF drop rules already skip them. Ads brand safety did not.

AdsBrandSafetyVfHydrator compute_verdict used type presence only. A lapsed NSFA_COMMUNITY_NOTE (or other TTL label) kept the post MediumRisk, so ads would not sit next to a post whose note or label had already expired.

## Fix

Skip labels whose expires_at_msec is at or before now before scoring the verdict. Missing expiry stays permanent.

No change to write paths, TTLs, GetSafetyLabels, or VF drop rules.

## Twin

VF expiry fence is already covered separately. Community Notes have no VF drop rule in this dump; ads adjacency is the ranking sink for NSFA_COMMUNITY_NOTE.


Pitchfork-and-Torch:
Duplicate of #124 (same fix, agent already opened upstream).

PR #127, opened 2026-09-07
home-mixer: stop ads brand-safety from demoting organic rank

Pitchfork-and-Torch:
## Bug

Ads brand-safety was leaking into organic For You rank.

Two writers put MediumRisk on organics that were not MediumRisk from tweet labels:

1. Missed ads-VF hydration left `brand_safety_verdict` unset. Scored-post converters then defaulted that to MediumRisk.
2. `nsfw_author_ads` is ads inventory. AdsBrandSafetyVfHydrator wrote that onto the shared verdict the blender uses for organics.

Default For You blender (`partition_organic`) treats MediumRisk as avoid and pulls those posts out of ranked order when ads are placed.

## Fix

- Missing verdict serializes as Unspecified. Unspecified is not avoid in `has_avoid`, so rank is kept.
- Ads-only `nsfw_author_ads` no longer rewrites the organic verdict.
- Real MediumRisk from tweet labels still stays off ad neighbors.

## Rebase

Rebased `cursor/ads-organic-verdict-leak-e878` onto latest `xai-org/x-algorithm` `main` (`902a06f`). The previous head sat on the older fork tree, so GitHub reported CONFLICTING / dirty.

Conflict resolution keeps all three intents:

1. Expired safety labels are ignored in ads brand safety (`is_active_label` / `compute_verdict_at`) — same class as #124.
2. Missing ads verdict is Unspecified, not MediumRisk; `nsfw_author_ads` does not rewrite the organic verdict — #127.
3. Gizmoduck uses origin-author size, NSFW OR across poster/original/quoted, and fail-closed on store miss — #125, because that file conflicted.

Not a TimelineHome VF change.

### DIFF
--- a/home-mixer/ads/tests/partition_organic_blender_tests.rs
+++ b/home-mixer/ads/tests/partition_organic_blender_tests.rs
@@ -110,12 +110,75 @@ fn ad_bsr_levels(items: &[FeedItem]) -> Vec<BrandSafetyRiskLevel> {
         .collect()
 }
 
+fn organic_ids(items: &[FeedItem]) -> Vec<u64> {
+    items
+        .iter()
+        .filter_map(|item| match &item.item {
+            Some(feed_item::Item::Post(p)) => Some(p.tweet_id),
+            _ => None,
+        })
+        .collect()
+}
+
+#[test]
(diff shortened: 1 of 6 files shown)


Pitchfork-and-Torch:

PR #128, opened 2026-09-07
Stop For You from stamping OON before TES fills author_id

Pitchfork-and-Torch:
## Bug

For You (phoenix_candidate_pipeline) runs InNetworkCandidateHydrator before CoreDataCandidateHydrator. InNetwork reads author_id and writes in_network once.

TweetMixer often ships author_id = 0. InNetwork treats 0 as unfollowed and stamps in_network = Some(false). TES later writes the real author. The flag stays false.

VF then fetches at TimelineHomeRecommendations. Recs-only Drop rules fire (DO_NOT_AMPLIFY, SPAM_HIGH_RECALL, NSFW_HIGH_RECALL, MALICIOUS_URL, FOSNR_ABUSE_INSULTS, and the rest of the Recs table). The same followed author on Home is Allow (or Interstitial for NSFW HP / gore / card).

phoenix_scores_pipeline already runs CoreData, then InNetwork.

## Fix

Run CoreData, then InNetwork, same as the scores pipeline.

This is not rewriting Recs-only policy. Not #115 / #116 / #119.

PR #131, opened 2026-09-07
Stop CoreData from keeping a stale source author_id after TES

Pitchfork-and-Torch:
## Bug

#128 runs InNetwork after CoreData so TES can fill `author_id`. CoreData `update` only wrote `author_id` when the source field was already 0.

Phoenix (and Thunder) ship a nonzero `author_id`. TES still runs (text is empty, so `already_hydrated` is false). The TES author is discarded. InNetwork stamps the source id.

A followed TES author is then fetched at `TimelineHomeRecommendations`. Recs-only Drop fires for tweet `NSFW_HIGH_PRECISION`, `GORE_AND_VIOLENCE_HIGH_PRECISION`, and `NSFW_CARD_IMAGE`. The same labels on Home are Interstitial (`NSFW_MEDIA_INTERSTITIALS`).

- Entry: `CoreDataCandidateHydrator::update` fill-if-zero
- Sink: `VFCandidateHydrator` (`in_network.unwrap_or(false)` → Recs) then `VFFilter`
- Break: TES author thrown away when the source already set a nonzero id
- Viewer effect: a followee's NSFW HP / gore / card post retrieved via Phoenix is hard-dropped on For You
- Twin: `retweeted_tweet_id` / `retweeted_user_id` from TES already overwrite the source

This is not rewriting Recs-only policy. This is not #128 (order). This is not #119 (RT wrapper primary id). This is not #115 (Following quotes).

## Fix

When TES returns a nonzero author, write it. TES miss still keeps the source author.

## Tests

- Source `author_id = 0`, TES `42` → `42`
- Source `99`, TES `42` → `42`
- TES miss (`0`) keeps source
- TES `42` (followed) then InNetwork → `in_network = true`
- Stale source `99` without TES overwrite still stamps OON

Standalone decision-table harness (same predicate): 6 assertions passed.

## Leftover

`already_hydrated` still skips TES when source `author_id != 0` and text is nonempty. `InNetworkCandidateHydrator.enable` is still `!has_cached_posts`.

### DIFF
--- a/home-mixer/candidate_hydrators/core_data_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/core_data_candidate_hydrator.rs
@@ -118,15 +118,24 @@ impl CachedHydrator<ScoredPostsQuery, PostCandidate> for CoreDataCandidateHydrat
     }
 
     fn update(&self, candidate: &mut PostCandidate, hydrated: PostCandidate) {
-        if candidate.author_id == 0 && hydrated.author_id != 0 {
-            candidate.author_id = hydrated.author_id;
-        }
-        candidate.retweeted_user_id = hydrated.retweeted_user_id;
-        candidate.retweeted_tweet_id = hydrated.retweeted_tweet_id;
-        candidate.in_reply_to_tweet_id = hydrated.in_reply_to_tweet_id;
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

PR #132, opened 2026-09-07
Stop CoreData from skipping TES on NightOwl replies and retweets

Pitchfork-and-Torch:
## Bug

Latest Following (NightOwl) pre-fills author_id and tweet_text. CoreDataCandidateHydrator.already_hydrated treated that as done and skipped TES.

NightOwl does not set ancestor_users or retweeted_user_id. SelfReplyChainFilter fail-opens when ancestor_users is empty, so self-replies aimed at an unfollowed user stay in Following.

## Fix

Call TES when a reply still has empty ancestor_users, or a retweet still has no retweeted_user_id. Originals with author and text still skip TES.

On TES miss, keep the source fields.

Not #94 (TES miss wiping Thunder/Phoenix ids). Not #99 (For You self-reply chains). Related to but distinct from CoreData stale author overwrite (#131).


Pitchfork-and-Torch:

PR #133, opened 2026-09-07
Stop treating gizmoduck age 0 as a confirmed underage viewer

Pitchfork-and-Torch:
## Bug

Gizmoduck age_in_years: Some(0) is a missing-birthday sentinel. Visibility filtering mapped it to ViewerAge::Known(0). viewer_is_underage treated any Known(age) < 18 as a child, so SensitiveViewerUnderageDropRule hard-dropped sensitive media worldwide for those accounts.

RPC error already fails open as Unknown. Existing users with no usable birthday should be NotStated (jurisdiction-scoped via NSFW_GATING_COUNTRIES). Home-mixer and Phoenix already treat 0 as missing. VF did not.

## Fix

Some(age) if age > 0 → Known. Else existing user → NotStated. Else → Unknown. Underage is logged-in + age in 1..18.

Age gates stay. Logged-out, confirmed underage, and no-stated-age-in-gating-country drops are unchanged.

Also: BidirectionalFollowHydrator now runs after CoreData and skips author_id 0 (leftover from #128). InNetwork order from #128 is unchanged.

### DIFF
--- a/home-mixer/candidate_hydrators/bidirectional_follow_hydrator.rs
+++ b/home-mixer/candidate_hydrators/bidirectional_follow_hydrator.rs
@@ -30,10 +30,12 @@ impl Hydrator<ScoredPostsQuery, PostCandidate> for BidirectionalFollowHydrator {
             .collect();
         let check_all_authors = query.params.get(EnableAllAuthorFollowHydration);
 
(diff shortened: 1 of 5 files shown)


Pitchfork-and-Torch:

PR #135, opened 2026-09-07
Recompute in_network on cached For You posts

Pitchfork-and-Torch:
## Bug

`InNetworkCandidateHydrator.enable` is `!has_cached_posts`.

`FollowedUserIdsQueryHydrator` still runs on a cache hit. The follow list is current. The hydrator is a set lookup on `author_id`. Skipping it keeps the `in_network` bit written up to 180s earlier.

`VFCandidateHydrator` buckets on `in_network.unwrap_or(false)`. Stale `Some(true)` after unfollow is fetched at `TimelineHome`. Recs-only Drop does not run (`DO_NOT_AMPLIFY`, `SPAM_HIGH_RECALL`, `NSFW_HIGH_RECALL`, civic-integrity, and the rest of the Recs table). `OONRetweetReplyFilter` and `OONNsfwSimclustersFilter` also need `Some(false)`.

This is the leftover called out on #131. Not #128 (TES-before-InNetwork order). Not #117 (QuoteHydrator graph miss).

## Fix

Always enable. Recompute from the live follow list. `update` overwrites the cached bit.

## Tests

- `enable` is true on cache hit and miss
- cache-hit hydrate + update: newly followed → `true`, unfollowed → `false`, self → `true`

Standalone decision-table harness (same enable + VF bucket predicate): 6 assertions passed.

### DIFF
--- a/home-mixer/candidate_hydrators/in_network_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/in_network_candidate_hydrator.rs
@@ -8,8 +8,10 @@ pub struct InNetworkCandidateHydrator;
 
 #[async_trait]
 impl Hydrator<ScoredPostsQuery, PostCandidate> for InNetworkCandidateHydrator {
-    fn enable(&self, query: &ScoredPostsQuery) -> bool {
-        !query.has_cached_posts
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

Pitchfork-and-Torch:

PR #136, opened 2026-09-07
Stop InNetwork from skipping cache hits and leaving a stale VF stamp

Pitchfork-and-Torch:
## Bug

#128–#132 leftover. `InNetworkCandidateHydrator.enable` was `!has_cached_posts`. #131 called that out and left it.

`FollowedUserIdsQueryHydrator` still runs on a cache hit. InNetwork is a local HashSet lookup. Redis keeps the old `in_network` for 180s. `VFCandidateHydrator` still runs and buckets `TimelineHome` vs `TimelineHomeRecommendations` from that stamp.

- Unfollow a DoNotAmplify / NSFW HP author: cache still says Home. Recs Drop does not fire.
- Follow someone: cache still says OON. Recs Drop fires on a followee.
- Cache miss that stamped OON on `author_id = 0` (#128), then TES filled the author: the Redis row keeps `in_network = false`. Cache hit never restamps.

CoreData stays skipped. TES is an RPC. Tweet `author_id` does not change. #132 already covers NightOwl `already_hydrated`. Gizmoduck stays skipped. VF still applies NSFW on the restamped safety level.

Not #128 (order). Not #131 (TES overwrite). Not #132 (`already_hydrated`).

## Fix

One file, `home-mixer/candidate_hydrators/in_network_candidate_hydrator.rs`.

`enable` is always true.

## Tests

- `enable` is true on `has_cached_posts`
- Cached Home stamp after unfollow becomes OON
- Cached Recs stamp after follow becomes Home
- Followed / unfollowed miss path unchanged

`cargo test` cannot run. Public dump has no Home Mixer manifest. Same control flow compiled and run as a rustc model: 6/6 passed (old gate keeps the stale stamp; new gate restamps).

### DIFF
--- a/home-mixer/candidate_hydrators/in_network_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/in_network_candidate_hydrator.rs
@@ -8,8 +8,11 @@ pub struct InNetworkCandidateHydrator;
 
 #[async_trait]
 impl Hydrator<ScoredPostsQuery, PostCandidate> for InNetworkCandidateHydrator {
-    fn enable(&self, query: &ScoredPostsQuery) -> bool {
-        !query.has_cached_posts
+    // Always on. FollowedUserIdsQueryHydrator is request-fresh and this is a
+    // local HashSet lookup. Skipping cache hits leaves Redis in_network (TTL
+    // 180s) in place; VF then buckets TimelineHome vs Recs from that stamp.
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:
Duplicate of #135 (same enable=true on cache hits). Prefer #135 as the ship vehicle.

PR #134, opened 2026-09-07
Fail closed when TES NSFW, takedown, nullcast, or media reads error

Pitchfork-and-Torch:
## Wave 2 check

Quote / RT / ancestor `Action::Interstitial` does not set `drop_ancillary_posts`. Verified.

`should_drop_reason` only matches `Action::Drop`. `AncillaryVFFilter` reads that flag. Home NSFW rules emit Interstitial; Recs has Drop twins (`NSFW_HIGH_PRECISION_DROP`, etc.). Interstitial means keep-and-warn, not hard-drop. Not rewritten.

This is not #115 (Following quotes at Home), #117 (QuoteHydrator TES), #118 (quoted id on RTs), #119 (RT primary looked up by wrapper id), or #121 (VF `Err` / missing key). #121 leftover text called this Interstitial≠Drop path out and said TES-flag miss is separate.

## Bug

TES tweet-flag RPCs (`nsfw_user`, `nsfw_admin`, `takedown_reasons`, `nullcast`, `media`) already distinguish Found / NotFound / Failed. `build_tweet_features` then does `get(...).unwrap_or(false)` and `get_or_default`.

Failed and NotFound both become “flag unset”. NSFW interstitial / Recs Drop, legal takedown, nullcast, and DMCA/geo rules never see the label.

- Entry: `TesHydrator::hydrate_tweets` → `TweetHydration`
- Sink: `NsfwAuthorInterstitialRule`, tweet NSFW Recs drops, legal takedown, `NullcastedTweetDropRule`, Recs DMCA/geo
- Break: Failed TES flag read assembled as false
- Viewer effect: NSFW or taken-down post (or a quote/reply of one) serves as a normal card
- Twin: #121 fail-closes a missing VF map key. TES flags were the leftover that issue called out.

Genuine `NotFound` / `Ok(None)` (no flag) is unchanged.

## Fix

If any of those five lookups is Failed for a tweet id, drop that candidate from the hydrated set. `FilterTweets` already maps a missing hydrated candidate to `Verdict::unresolved_author` (Drop / `UnspecifiedReason`). Mixer `VFFilter` and `should_drop_ancillary` already drop that reason.

Do not treat Interstitial as Drop.

## Tests

- Found false / NotFound / empty batch → not a safety failure
- Failed nsfw_user, takedown_reasons, nullcast, or media → safety failure
- `retain_candidates_with_usable_tes_flags` drops only the Failed id
- `Hydrated::is_failed` is true only for Failed

[...]

PR #137, opened 2026-09-07
Fail closed when socialgraph mute/block relationship reads fail

Pitchfork-and-Torch:
## Wave 3 check

#134 leftover: socialgraph relationship miss still fail-opens as not muted / not blocked. Verified on `xai-org/main` (`902a06f`).

This is not #134 (TES NSFW / takedown / nullcast / media flags). This is not #121 (mixer VF `Err` / missing key). Those stay on their own PRs.

## Bug

`SocialgraphHydrator` already marks an empty or failed Flock read as `Hydrated::Failed`. `assemble` then does `relationships.get_or_default`, so Failed becomes all-false. `ViewerMutesAuthorRule` / `ViewerBlocksAuthorRule` / `MutedRetweetsRule` never fire.

Latest Following never loads mixer mute/block lists (`reverse_chron_posts_pipeline` query hydrators are empty). For You list hydrators return `Err` on RPC failure; `hydrate_query` ignores that and leaves empty lists. In both cases VF is the mute/block sink. The sink was assembling Failed as not-muted / not-blocked.

Mixer `BlockedByHydrator` / `FollowingBlockedByHydrator` returned `Err` on `check_blocked_by` failure. `update_all` skips `Err`, so `author_blocks_viewer` stays `None`. `AuthorSocialgraphFilter` does `unwrap_or(false)`. VF does not check author-blocks-viewer.

- Entry: `SocialgraphHydrator::hydrate` → `CandidateFeatures::assemble`; mixer `BlockedByHydrator` / `FollowingBlockedByHydrator`
- Sink: VF mute/block/mute-retweet rules; mixer `AuthorSocialgraphFilter`
- Break: Failed relationship assembled as false; mixer graph `Err` written as not-blocked
- Viewer effect: a muted or blocked author (or an author who blocked the viewer) still serves when the graph RPC fails
- Twin: #134 drops a candidate whose TES safety-flag read Failed. Same retain → `unresolved_author` path.

Genuine `Found` (including all-false, nobody muted/blocked) is unchanged. Logged-out viewers are `Found(default)`, not Failed.

## Fix

If the relationship lookup is Failed for a tweet id, drop that candidate from the hydrated set. `FilterTweets` already maps a missing hydrated candidate to `Verdict::unresolved_author`.

On mixer `check_blocked_by` `Err`, write `author_blocks_viewer: Some(true)` (and quoted `Some(true)` when a quoted author is present) so `update_all` applies it and the filter drops.

## Tests

- Failed relationship dropped; Found default kept
- Found mute/block still assembled (rules drop)
- Missing relationship key dropped
- Assemble-without-retain still defaults Failed to false (documents the hole)

Standalone decision-table harness (same retain / BlockedBy / filter arms): 15 assertions passed.

`cargo test` cannot run. Public dump has no visibility-filtering / Home Mixer manifest.

## Leftover

`QuoteHydrator::get_blocked_by` still `unwrap_or_default()` (#117). Partial Flock result slots still default missing mute/block graphs to empty. Mixer mute/block *list* miss on For You still leaves empty lists; Latest Following never loads those lists. VF Failed is the sink for viewer mute/block after this change.

Fork pointer (do not merge): https://github.com/Pitchfork-and-Torch/x-algorithm/pull/27

### DIFF
--- a/home-mixer/candidate_hydrators/blocked_by_hydrator.rs
+++ b/home-mixer/candidate_hydrators/blocked_by_hydrator.rs
@@ -34,9 +34,19 @@ impl Hydrator<ScoredPostsQuery, PostCandidate> for BlockedByHydrator {
             .await
         {
             Ok(ids) => ids,
-            Err(e) => {
-                let err_msg = e.to_string();
-                return candidates.iter().map(|_| Err(err_msg.clone())).collect();
+            Err(_) => {
+                // Err is skipped by update_all, leaving author_blocks_viewer
+                // None. AuthorSocialgraphFilter then unwrap_or(false).
+                // Write Some(true) so the filter drops.
+                return candidates
+                    .iter()
+                    .map(|_| {
+                        Ok(PostCandidate {
+                            author_blocks_viewer: Some(true),
+                            ..Default::default()
+                        })
+                    })
+                    .collect();
             }
         };
         candidates
--- a/home-mixer/candidate_hydrators/following_blocked_by_hydrator.rs
+++ b/home-mixer/candidate_hydrators/following_blocked_by_hydrator.rs
@@ -33,9 +33,19 @@ impl Hydrator<ScoredPostsQuery, PostCandidate> for FollowingBlockedByHydrator {
             .await
         {
             Ok(ids) => ids,
-            Err(e) => {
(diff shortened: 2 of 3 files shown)


Pitchfork-and-Torch:

PR #138, opened 2026-09-07
Stop Thunder from deleting quotes on QuotedTweetDeleteEvent

Pitchfork-and-Torch:
## Bug

QuotedTweetDeleteEvent is a quote-card update. The quoting post still exists. Thunder treated quoting_tweet_id as a TweetDelete and removed the quote from the in-network store.

TweetDeleteEvent already removes the tweet that was actually deleted (the original, or a quote the author deleted).

## Proof

- Entry: thunder/kafka/tweet_events_listener.rs (feeder InNetworkEvent deletes)
- Sink: PostStore.mark_as_deleted -> GetInNetworkPosts
- Break: QuotedTweetDeleteEvent used quoting_tweet_id as a delete
- Viewer effect: a followee quote vanished from Following / For You in-network when the quoted original was deleted
- Twin: TweetDeleteEvent already deletes the real tweet; QuotedTweetTakedownEvent was already ignored

## Fix

Ignore QuotedTweetDeleteEvent for in-network deletes. Keep TweetDeleteEvent retention as-is.

## Tests

Standalone decision table (same predicate): 4/4 passed.

- TweetDelete within retention still deletes that tweet
- TweetDelete past retention stays ignored
- QuotedTweetDelete does not delete the quoting post
- QuotedTweetTakedown still does not delete

`cargo test` cannot run. Public dump has no Thunder Cargo.toml.

## Notes

Not TES id wipe (#94). Not QuoteHydrator (#117/#118). Not VF quote scoring (#115/#119).

### DIFF
--- a/thunder/kafka/tweet_events_listener.rs
+++ b/thunder/kafka/tweet_events_listener.rs
@@ -52,6 +52,30 @@ async fn monitor_partition_lag(
     }
 }
 
+/// TweetDeleteEvent removes the tweet that was actually deleted.
+/// QuotedTweetDeleteEvent is a quote-card update: the quoting post still
+/// exists. Using quoting_tweet_id here wipes live quotes from Thunder when
+/// the quoted original is deleted. TweetDeleteEvent already covers both
+/// the original and a quote that the author deleted.
+fn in_network_delete_post_id(
+    data: &TweetEventData,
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #104, opened 2026-09-04
scarecrow: make stale NSFW_CARD_IMAGE URL verdict cleanup reachable

Pitchfork-and-Torch:
## Problem

Rule 7429 writes a 7-day NSFW_CARD_IMAGE verdict on near-perfect NSFW card images. Its cleanup branch was meant to delete that verdict when a later score is clean, but it could never run: contradictory guards, inverted age check, and wrong time units.

False-positive card URL labels therefore lasted the full TTL, interstitialing linked posts and dropping them from out-of-network recommendations with no automatic correction.

## Change

Make the cleanup branch reachable and arithmetically correct. Leave the write path untouched.

Fork PR: https://github.com/Pitchfork-and-Torch/x-algorithm/pull/8

PR #139, opened 2026-09-07
Fail closed when VF safety-label lookup is Err or missing

Pitchfork-and-Torch:
## Bug

`SafetyLabelHydrator` turns a failed or omitted RTF / Manhattan label read into `SafetyLabelMap::default()`. NSFW_CARD_IMAGE drop and interstitial only check type presence, so an empty map is Allow.

`Ok(empty)` is a confirmed unlabeled tweet. That path is unchanged.

This is not #104 (bot 7429 cleanup reachability). This is not #107 (GetTweetLabelInfoFromURL treating Failed as unlabeled on the write side). This is the VF consumer of those same RTF labels.

## Five-line proof

- Entry: `SafetyLabelHydrator::hydrate` (`and_then(Result::ok)` used to collapse Err/missing into empty)
- Sink: `FilterTweets` omit-from-hydrate -> `Verdict::unresolved_author` Drop
- Break: lookup Err and omitted id both became unlabeled, so NSFW_CARD_IMAGE never fired
- Viewer effect: a labeled URL card / NSFW post whose label store missed still served
- Twin: ads brand-safety already fail-closes `failed_ids`; GetSafetyLabels RPC already returns them separately

## Change

Track `failed_ids`. Do not insert a default empty map on Err or missing key. `retain_candidates_with_usable_safety_labels` omits those ids before assemble. Confirmed empty maps stay unlabeled and serve.

## Tests

- Lookup Err is failed (no default empty map)
- Confirmed empty map is not a failure
- Retain drops only failed ids
- Successful labeled / unlabeled reads unchanged

Standalone decision-table harness (same match arms): 12 assertions passed.

`cargo test` cannot run. Public dump has no VF crate manifest.

Fork PR: none

### DIFF
--- a/visibility-filtering/hydration/mod.rs
+++ b/visibility-filtering/hydration/mod.rs
@@ -21,7 +21,9 @@ use batch::TweetHydrationBatch;
 use exclusive_content_hydrator::ExclusiveContentHydrator;
 use fallback_cache::FallbackCache;
 use gizmoduck_hydrator::GizmoduckAuthorHydrator;
-use safety_label_hydrator::{SafetyLabelHydration, SafetyLabelHydrator};
+use safety_label_hydrator::{
+    retain_candidates_with_usable_safety_labels, SafetyLabelHydration, SafetyLabelHydrator,
+};
 use socialgraph_hydrator::SocialgraphHydrator;
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:

PR #129, opened 2026-09-07
Stop IneligibleSubscriptionFilter from dropping exclusive posts on a subscription-list miss

Pitchfork-and-Torch:
## Bug

For You `IneligibleSubscriptionFilter` drops exclusive posts when the subscription list was never loaded, and when the viewer is the conversation author.

`SubscribedUserIdsQueryHydrator` errors are ignored by `hydrate_query`. The list stays empty. The mixer filter treated empty as "subscribed to nobody" and dropped every `subscription_author_id: Some(...)` before scoring.

VF `DropExclusiveTweetContentRule` allows the conversation author. The mixer filter did not.

- Entry: Phoenix `IneligibleSubscriptionFilter` (`phoenix_candidate_pipeline.rs`)
- Sink: pre-scoring drop of exclusive / Super Follow posts
- Break: socialgraph miss → empty list → drop all exclusive; conversation author never exempted
- Viewer effect: paying subscribers lose exclusive posts on a list-RPC miss; conversation authors lose exclusive replies in For You
- Twin: VF still decides exclusive access when the mixer list is missing

This is not #96–#120 (VF, mute/block, quote, scarecrow, geo). Brazil 2026 filter is not in this snapshot.

## Fix

Mark the subscription list hydrated only after a successful socialgraph read. Keep exclusive posts when the list was not loaded. Keep exclusive posts when the viewer is the conversation author.

A hydrated empty list still drops exclusive posts for everyone except the conversation author.

## Tests

Filter unit tests cover hydrator-miss keep, conversation-author keep, subscribed keep, unsubscribed drop, and hydrated-empty drop. Hydrator tests cover success marking the list ready and errors leaving it unready.

cargo: cannot run the Home Mixer crate here. Public dump has no Home Mixer manifest. Standalone keep-rule check: 6/6 passed.

PR #141, opened 2026-09-07
Fail closed when exclusive TES lookup is Err or missing

Pitchfork-and-Torch:
## Bug

TES `get_exclusive_controls` timeout or per-id `Err` was collapsed to `exclusive_content = None`. `DropExclusiveTweetContentRule` treats `None` as not exclusive and Allows.

Mixer `SubscriptionHydrator` returned `Err` on `get_subscription_author_ids` miss. `CachedHydrator::update_all` skips that write, so `subscription_author_id` stays `None`. `IneligibleSubscriptionFilter` keeps `None`.

This is not #129 (viewer subscription-list miss starve). #129 keeps exclusive posts when the list was not loaded and relies on VF. VF was fail-opening the same TES exclusive read.

- Entry: `ExclusiveContentHydrator` (`timed_rpc` empty map on timeout; `Err`/`None` both become not exclusive). Mixer `SubscriptionHydrator::hydrate_from_client`.
- Sink: `DropExclusiveTweetContentRule` (`!is_exclusive() => Allow`). For You also `IneligibleSubscriptionFilter` (`None => keep`). Both sit in `phoenix_candidate_pipeline` / `reverse_chron_posts_pipeline` via `VFFilter`.
- Break: TES exclusive RPC error or omitted id assembled as not-exclusive.
- Viewer effect: a Super Follow / exclusive post ranks as a normal card for non-subscribers on For You and Latest Following.
- Twin: Genuine `Ok(None)` (tweet is not exclusive) is unchanged. Super-follow graph miss still fail-closes as not subscribed.

## Fix

VF: run exclusive TES through `timed_results` so Found / NotFound / Failed stay distinct. Failed or missing id sets `exclusive_hydration_failed`. The exclusive rule Drops that candidate with `ExclusiveTweet`.

Mixer: write `subscription_lookup_failed` on TES `Err` or missing key so `update_all` applies it. Cache stores Failed separately from `Resolved(None)`. The mixer filter drops a failed lookup.

## Tests

- `exclusive_slot`: Found exclusive, NotFound, Err, missing key, timeout batch
- Assemble stamps `exclusive_hydration_failed` on TES Err; NotFound stays not exclusive
- `DropExclusiveTweetContentRule` drops Failed for viewer, author, and logged-out
- Golden corpus case `exclusive_tes_lookup_failed_drops`
- Mixer TES Err/missing fail-closed; `Ok(None)` stays not exclusive
- Mixer filter drops failed lookup; keeps genuine not-exclusive
- Cache round-trip preserves Failed vs Resolved(None)

Standalone decision-table harness (same VF + mixer arms): 12/12 passed.

`cargo test` cannot run. Public dump has no visibility-filtering / Home Mixer manifest.

## Leftover

Gizmoduck author NSFW miss remains #125. Socialgraph mute/block Failed is #137. Interstitial on an ancillary quote/RT/ancestor still does not set `drop_ancillary_posts` (keep-and-warn; #134 left it).

Fork PR: none

### DIFF
--- a/home-mixer/candidate_hydrators/subscription_hydrator.rs
+++ b/home-mixer/candidate_hydrators/subscription_hydrator.rs
@@ -6,9 +6,15 @@ use tonic::async_trait;
 use xai_candidate_pipeline::component_library::utils::{default_quick_cache, QuickCache};
 use xai_candidate_pipeline::hydrator::{CacheStore, CachedHydrator};
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum SubscriptionAuthorHydration {
+    Resolved(Option<u64>),
+    Failed,
+}
+
 pub struct SubscriptionHydrator {
     pub tes_client: Arc<dyn TESClient + Send + Sync>,
-    pub cache: QuickCache<u64, Option<u64>>,
+    pub cache: QuickCache<u64, SubscriptionAuthorHydration>,
 }
 
 impl SubscriptionHydrator {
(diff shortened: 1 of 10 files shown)


Pitchfork-and-Torch:

PR #142, opened 2026-09-07
Stop RankingScorer from ranking posts with no Phoenix heads

Pitchfork-and-Torch:
## Bug

Phoenix lookup miss, missing `scoring_sequence`, or overflow past `PHOENIX_CLIENT_MAX_CANDIDATES` leaves every Phoenix head `None`. `RankingScorer::apply` treated that as 0, then `offset_score` wrote `Some(NEGATIVE_SCORES_OFFSET)`. `TopKScoreSelector` only drops `score == None`. Those posts still ranked.

`Some(0.0)` is a real Phoenix prediction and still ranks.

This is not #3 (ranker xDS). This is not #31 (author-size IPS). This is not a taste weight tweak.

## Five-line proof

- Entry: `PhoenixScorer` / `predictions.candidate_scores` leave all heads None on miss
- Sink: `PhoenixCandidatePipeline` scorers `[PhoenixScorer, RankingScorer, VMRanker]` then `TopKScoreSelector` (`score.unwrap_or(NEG_INFINITY)`)
- Break: `apply(None) = 0` plus offset wrote a finite rank for an unscored post
- Viewer effect: For You could serve a tweet Phoenix never scored, including overflow and no-sequence requests
- Twin: scorer `update_all` already skips `Err`; prediction failure already returned `Err`

## Change

- `RankingScorer` and `PhoenixScoresRankingScorer`: no heads -> `score` / `weighted_score` stay `None`
- `PhoenixScorer`: missing `scoring_sequence` returns `Err` so `update_all` does not stamp empty scores or `last_scored_at_ms`

## Tests

- Default heads: score unset
- `Some(0.0)` favorite: still ranks
- Mixed slate: only the unscored candidate is unset
- Missing sequence: Err, not Ok(default)

`cargo test` cannot run. Public dump has no home-mixer crate manifest.

Fork PR: none

### DIFF
--- a/home-mixer/scorers/phoenix_scorer.rs
+++ b/home-mixer/scorers/phoenix_scorer.rs
@@ -21,6 +21,18 @@ pub struct PhoenixScorer {
 }
 
 impl PhoenixScorer {
+    fn refuse_without_sequence(
+        query: &ScoredPostsQuery,
+        n: usize,
+    ) -> Option<Vec<Result<PostCandidate, String>>> {
+        query.scoring_sequence.is_none().then(|| {
+            vec![
+                Err("Phoenix scoring_sequence missing; refusing empty scores".to_string());
+                n
+            ]
+        })
+    }
+
     fn resolve_cluster(query: &ScoredPostsQuery) -> PhoenixCluster {
         let configured_cluster =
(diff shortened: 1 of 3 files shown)


Pitchfork-and-Torch:

PR #143, opened 2026-09-07
Drop quotes, retweets, and replies whose ancillary VF verdict is Interstitial

Pitchfork-and-Torch:
## Bug

#134 leftover. Quote / RT / ancestor `Action::Interstitial` does not set `drop_ancillary_posts`. Verified on `xai-org/main` (`902a06f`).

`should_drop_reason` only matched `Action::Drop`. Home NSFW / gore / card / nsfw-author rules emit Interstitial. Recs has Drop twins, but RT sources are fetched at `TimelineHome`, and a dual-role id is overwritten by the Home Interstitial. `AncillaryVFFilter` reads the flag. The wrapper `visibility_reason` is the primary Allow. Mixer serializes only that. This repo does not draw an interstitial.

This is not #115 (Following quotes at Home vs Recs). Not #119 (RT primary looked up by wrapper id). Not #121 (VF `Err` / missing key). #121 leftover text called this Interstitial≠Drop path out. #134 left it as keep-and-warn. Keep-and-warn is only real on the primary card.

## Five-line proof

- Entry: `VFCandidateHydrator::should_drop_reason` (quote / RT / ancestor lookups in `should_drop_ancillary`)
- Sink: `AncillaryVFFilter` (`drop_ancillary_posts`); `candidates_to_scored_posts` writes only primary `visibility_reason`
- Break: `Action::Interstitial` on a child did not set the flag; wrapper stayed Allow
- Viewer effect: a followee RT (or quote/reply) of NSFW / gore serves as a normal For You / Following card with the media embed and no warn
- Twin: Recs Drop twins already drop the same labels when the child is fetched at `TimelineHomeRecommendations`. Primary Interstitial still keep-and-warn via `VFFilter` / `visibility_reason`

## Change

Treat `Action::Interstitial` like `Action::Drop` for ancillary children only. Primary Interstitial is unchanged.

## Tests

- Quote / RT / ancestor Interstitial → `drop_ancillary_posts`
- Tombstoned ancestor Interstitial still skipped
- Ancillary Allow keeps; ancillary Drop still drops
- Primary Interstitial with no child does not set the flag

Standalone decision-table harness (same match arms): 9/9 passed.

Fork PR: none

### DIFF
--- a/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
@@ -176,9 +176,132 @@ pub(crate) fn should_drop_ancillary(
 
 fn should_drop_reason(reason: &FilteredReason) -> bool {
     match reason {
-        FilteredReason::SafetyResult(safety_result) => {
-            matches!(safety_result.action, Action::Drop(_))
-        }
-        _ => true, 
+        FilteredReason::SafetyResult(safety_result) => matches!(
+            safety_result.action,
+            // Keep-and-warn is only wired for the primary card (`visibility_reason`).
+            // An ancillary Interstitial is discarded, so the wrapper would serve
+            // NSFW/gore as a normal Allow embed. Drop the quote / RT / reply instead.
+            Action::Drop(_) | Action::Interstitial
+        ),
+        _ => true,
+    }
+}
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #144, opened 2026-09-07
Fail closed when gizmoduck user-label lookup is Err or missing

Pitchfork-and-Torch:
## Bug

`GizmoduckAuthorHydrator` keeps Failed and omitted gizmoduck reads as `Hydrated::Failed`. `CandidateFeatures::assemble` then calls `get_or_default`, which turns that into empty `AuthorFeatures` / empty `UserLabelSet`.

`SpamHighRecallUserLabelRule` (and sibling OON user-label drops) only check type presence. An empty set is Allow. A store error therefore lets a SpamHighRecall or LowQuality author keep ranking.

`Ok(None)` / `NotFound` is a confirmed unlabeled author. That path is unchanged.

This is not #103 (allowlist store). This is not #125 (home-mixer origin-author NSFW/size). This is not #139 (tweet RTF / Manhattan safety-label hydrator). This is the VF gizmoduck consumer of user labels.

## Five-line proof

- Entry: `GizmoduckAuthorHydrator` / `author_lookup_failed` (`get_or_default` used to collapse Err/missing into empty labels)
- Sink: `FilterTweets` omit-from-hydrate -> `Verdict::unresolved_author` Drop
- Break: gizmoduck Err and omitted id both became unlabeled, so `SpamHighRecallUserLabelRule` never fired
- Viewer effect: a SpamHighRecall / LowQuality author whose gizmoduck read missed still ranked
- Twin: #139 fail-closes tweet RTF labels; GetSafetyLabels already tracks `failed_ids`. This is gizmoduck user labels.

## Change

Track Failed / missing gizmoduck author reads. Do not assemble them as default empty labels. `retain_candidates_with_usable_author_features` omits those ids before assemble. Confirmed `NotFound` stays unlabeled and serves.

## Tests

- Lookup Err and omitted id are failed (not assembled as empty labels)
- Confirmed `NotFound` is not a failure
- Found SpamHighRecall / LowQuality labels stay usable
- Retain drops only failed ids
- Successful labeled / unlabeled reads unchanged

Standalone decision-table harness (same match arms): 11 assertions passed.

`cargo test` cannot run. Public dump has no VF crate manifest.

Fork PR: none

### DIFF
--- a/visibility-filtering/hydration/gizmoduck_hydrator.rs
+++ b/visibility-filtering/hydration/gizmoduck_hydrator.rs
@@ -1,5 +1,7 @@
 use crate::clients::gizmoduck_client::GizmoduckLookup;
-use crate::hydration::batch::{AuthorHydrationBatch, HydrationBatch, TweetHydrationBatch};
+use crate::hydration::batch::{
+    AuthorHydrationBatch, Hydrated, HydrationBatch, TweetHydrationBatch,
+};
 use crate::hydration::fallback_cache::FallbackCache;
 use crate::hydration::metrics::{record_batch_size, timed_results};
 use crate::hydration::{keyed_by_author, tweets_per_author};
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:

PR #145, opened 2026-09-07
Fail closed when TES edit-control lookup is Err or missing

Pitchfork-and-Torch:
TES `get_edit_control` timeout or per-id `Err` was collapsed to `edit_control = None`. `is_stale_tweet` treats `None` as current and Allows. `EditControl::Edit` with no `edit_control_initial` did the same.

This is not #134 (NSFW / takedown / nullcast / media TES flags). #134 leftover left edit-control out. Not #141 (exclusive TES). ConversationControl / TrustedFriends / MentionFilter have no TES hydrator on this dump (schema-only).

- Entry: `TesHydrator::hydrate_tweets` → `get_edit_control` (`timed_results` already Found / NotFound / Failed)
- Sink: `DropStaleTweetsRule` (`is_stale() && !is_retweet()`). Wired on TimelineHome and TimelineHomeRecommendations via `TES_HOME_DROPS` / `VFFilter`.
- Break: Failed TES edit-control RPC or omitted id assembled as not-stale. Missing `edit_control_initial` on the Edit variant also assembled as current.
- Viewer effect: a superseded / edited post ranks as the current card on For You and Latest Following.
- Twin: Genuine `Ok(None)` (never edited) is unchanged.

Stamp `edit_control_lookup_failed` when the edit-control slot is Failed. `is_stale_tweet` Drops Failed, empty edit chains, and Edit-without-initial. `DropStaleTweetsRule` already drops stale non-retweets.

- Assemble stamps lookup-failed on TES Err; NotFound stays not failed; Found current stays current
- Failed still stamps when core data is missing
- `is_stale_tweet` true for lookup-failed, empty chain, and Edit-without-initial
- `DropStaleTweetsRule` drops Failed for viewer, author, and logged-out
- Golden corpus: `stale_edit_control_lookup_failed_drops`, `stale_edit_control_missing_initial_drops`

Standalone decision-table harness (same Found / NotFound / Failed / missing-initial arms): 12/12 passed.

`cargo test` cannot run. Public dump has no visibility-filtering / Home Mixer manifest.

ConversationControl / TrustedFriends / MentionFilter are schema-only (thunder tweet/user). No TES RPC on this dump to hydrate an ACL. Exclusive TES fail-closed is #141. DropStaleTweetsRule still Allows stale retweets.

Fork PR: none

PR #147, opened 2026-09-07
Fail closed when TES edit_control lookup is Err

Pitchfork-and-Torch:
## Five-line proof

- Entry: `TesHydrator::hydrate_tweets` `get_edit_control` (`timed_results` Found / NotFound / Failed)
- Sink: omit-from-hydrate → `FilterTweets` `Verdict::unresolved_author` Drop
- Break: Failed assembled as `edit_control = None`; `is_stale_tweet` treats None as current
- Viewer effect: a superseded pre-edit tombstone (labeled old text) still serves
- Twin: #134 fail-closes TES nsfw / takedown / nullcast / media. `edit_control` was the leftover.

Genuine `NotFound` / `Ok(None)` (never edited) is unchanged.

`retain_candidates_with_usable_edit_control` drops Failed ids before assemble.

Tests: Found / NotFound / empty batch stay; Failed RPC and timeout drop; retain keeps only usable ids; assemble still collapses Failed to None.

Standalone decision-table harness (same match arms): 12/12 passed.

`cargo test` cannot run. Public dump has no visibility-filtering crate manifest.

Fork PR: none

### DIFF
--- a/visibility-filtering/hydration/mod.rs
+++ b/visibility-filtering/hydration/mod.rs
@@ -25,7 +25,7 @@ use safety_label_hydrator::{SafetyLabelHydration, SafetyLabelHydrator};
 use socialgraph_hydrator::SocialgraphHydrator;
 use std::collections::HashMap;
 use std::sync::Arc;
-use tes_hydrator::TesHydrator;
+use tes_hydrator::{retain_candidates_with_usable_edit_control, TesHydrator};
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:
Same class as #145 (TES edit_control fail-closed). Closing this twin so one class has one ship vehicle. Prefer #145 (rules + golden corpus + stale treatment).

PR #110, opened 2026-09-04
Stop applying country-scoped safety labels outside applicable_countries

Pitchfork-and-Torch:
## Bug
Tweet safety labels carry `applicable_countries` on the proto. SafetyLabelMap::from_proto_label_types keeps only the type key. Drop rules call has_safety_label, which is type presence only.

A Brazil-scoped DoNotAmplify or civic-integrity label therefore Drops a US viewer on TimelineHomeRecommendations. Geo-restricted media already reads the viewer country. Labels did not.

This is not expiry filtering (PR 106). Expired rows stay a separate class. Empty applicable_countries stays worldwide.

## Fix
Keep per-type country lists on SafetyLabelMap. TweetPredicates::has_safety_label uses applies(label, viewer.country_code). Empty or missing scope is worldwide. A scoped label does not apply when the viewer country is missing or outside the list.

## Proof
- Entry: SafetyLabel proto applicable_countries
- Sink: DoNotAmplifyOonDropRule / other tweet-label Drops via has_safety_label
- Break: from_proto_label_types discarded country scope
- Viewer effect: US For You dropped for a BR-only DNA/civic label
- Twin: DropTweetsWithGeoRestrictedMediaRule vs viewer country

## Tests
- empty_countries_is_worldwide
- scoped_label_applies_only_in_listed_country (fails on unmodified main)
- country_scoped_do_not_amplify_allows_out_of_country_oon
- country_scoped_do_not_amplify_drops_in_country_oon
- unit tests added; cargo test cannot run in the public dump

PR #148, opened 2026-09-07
Fail closed when withheld country list is missing or TES country-codes Err

Pitchfork-and-Torch:
## Bug

Country-scoped legal/local takedown allowed the post when the viewer country was missing. TES `get_takedown_country_codes` (`withheldInCountries`) was never fetched, so a country list that lived only on that RPC never reached the Drop rules. An empty reason country or a Failed country-codes read assembled as “not withheld”.

This is not #113 (account vs request country). This is not #110 (safety-label `applicable_countries`). This is not #134 (TES flag RPC Failed dropped from the hydrated set).

## Fix

Honor TES withheld-in-countries on `DropLegalTakendownPostRule`. Treat an empty reason country as worldwide (`xx`). If the withhold list is country-scoped and the viewer country is unknown, Drop. A Failed `get_takedown_country_codes` read assembles as `xx` so the legal rule fail-closes. Genuine NotFound / empty list stays un-withheld.

Author self-view is unchanged.

## Proof

- Entry: TES `get_takedown_country_codes` / `takedown_reasons.country_code`
- Sink: `DropLegalTakendownPostRule` / `DropLocalLawsTakendownPostRule`
- Break: country-codes RPC unused; unknown viewer + DE withhold Allowed; Failed list → empty → Allow
- Viewer effect: DE-withheld post (or a Failed country-list read) still serves on Home / For You
- Twin: media allow-list already drops unknown country (`xx`); legal withhold did not

## Tests

- `legal_takedown_country_scoped_drops_without_viewer_country` (fails on unmodified main)
- `legal_takedown_empty_country_is_worldwide`
- `withheld_in_countries_drops_matching_viewer`
- `withheld_in_countries_allows_other_country`
- `assemble_fail_closes_withheld_in_countries_err_as_worldwide`
- Standalone decision-table harness: 15 assertions passed

`cargo test` cannot run. Public dump has no visibility-filtering manifest.

Fork PR: none

### DIFF
--- a/visibility-filtering/hydration/tes_hydrator.rs
+++ b/visibility-filtering/hydration/tes_hydrator.rs
@@ -1,4 +1,4 @@
-use crate::hydration::batch::TweetHydrationBatch;
+use crate::hydration::batch::{Hydrated, TweetHydrationBatch};
 use crate::hydration::metrics::{record_batch_size, timed_keyed_rpc, timed_results};
 use crate::models::{
     CoreFeature, MediaFeature, NsfwFeature, TweetCandidateInput, TweetFeatures, TweetId,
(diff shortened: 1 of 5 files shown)


Pitchfork-and-Torch:

PR #149, opened 2026-09-07
Fail closed when VF gizmoduck author lookup is Err or missing

Pitchfork-and-Torch:
Gizmoduck author Err/missing is assembled via `get_or_default`. `NsfwAuthorInterstitial` and NSFW user-label drops only check presence, so a failed read looks unlabeled and the post ranks.

Drop candidates whose author lookup is Failed or missing. Confirmed NotFound is unchanged. `FilterTweets` already Drops omitted hydrations as `unresolved_author`.

This is not #125 (home-mixer gizmoduck). This is not #139 (tweet safety labels). This is not #134 (TES flags).


Pitchfork-and-Torch:
Same class as #144 (gizmoduck author/user-label lookup fail-closed on Err/missing). One class one PR. Prefer #144. Do not merge this twin.

PR #150, opened 2026-09-07
Stop Thunder from serving retweets of tombstoned originals

Pitchfork-and-Torch:
`mark_as_deleted` tombstones the original and drops its CompactPost; retweets of that id stay on the retweeter deque.
`get_posts_from_map` never checked `deleted_posts` for `source_post_id`, so a deleted in-network post kept serving as a retweet card.
Skip retweets whose source id is tombstoned. Live-source retweets are unchanged. Not #57 (sibling eviction), #59 (deque strip), or #138 (QuotedTweetDeleteEvent).
Test: delete original 100, retweet 200 is gone, retweet 201 of live 101 still serves. Standalone predicate 4/4.
thunder/ has no Cargo.toml in this snapshot.

### DIFF
--- a/thunder/posts/post_store.rs
+++ b/thunder/posts/post_store.rs
@@ -333,7 +333,20 @@ impl PostStore {
                 });
 
                 let light_post_iter = light_post_iter_1.filter(|post| {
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #151, opened 2026-09-07
Include quoted_tweet_id in seen/served related-post ids

Pitchfork-and-Torch:
`related_post_ids_iter` walked tweet/retweet/reply ids only, so seen/served filters missed quote-as-same-card.
QuoteHydrator fills `quoted_tweet_id` before PreviouslySeen, PreviouslySeenBackup, and PreviouslyServed.
Chain `quoted_tweet_id` next to retweet and reply. Leftover of #138, not a Thunder delete.
Tests: iterator includes the quote; backup filter drops a quote of an impressed original. Standalone 5/5.
home-mixer has no Cargo.toml in this snapshot.


Pitchfork-and-Torch:

PR #98, opened 2026-09-04
Honor exclude_videos on quotes of video posts

Pitchfork-and-Torch:
## Bug
VideoFilter drops when min_video_duration_ms is set on the primary card. QuoteHydrator already hydrates quoted_video_duration_ms, but only when EnableQuotedVqvDurationCheck is on (ranking VQV). exclude_videos clients still saw quotes of videos.

This is not muted-keyword quoted text (PR 96). Not mute-on-quote author id (PR 23).

## Fix
Fetch quoted video duration when exclude_videos is set. VideoFilter drops if either the primary or the quoted duration is present.

QuoteHydrator is skipped for cached posts (has_cached_posts). Cached quote duration is leftover if the cache payload lacks quoted_video_duration_ms.

## Proof
- Entry: QuoteHydrator get_min_video_durations
- Sink: VideoFilter when query.exclude_videos
- Break: filter ignored quoted_video_duration_ms; hydrator skipped duration unless VQV param
- Viewer effect: exclude_videos request still serves a quote of a video
- Twin: min_video_duration_ms on the primary card

## Tests
- test_removes_quotes_of_videos
- existing primary video tests still pass
- unit tests added; cargo test cannot run in the public dump

PR #152, opened 2026-09-07
Stop Latest Following from ignoring hide-replies and hide-links

Pitchfork-and-Torch:
## Bug

Latest Following copies `exclude_videos` (#120) and still ignores the rest of the viewer content-controls.

QueryBuilder never copied `hide_replies` / `hide_links` / `exclude_replies` / `exclude_retweets`. Night Owl always searched with `include:retweets` and never applied `-filter:replies` or `-filter:links`. Reverse-chron filters were RT-dedup, muted keyword, and self-reply. Hide-replies and hide-links were a no-op.

Ranked Following / For You already have `VideoFilter` for hide-videos. They had no reply/link/retweet content-control filter either; the same mixer filter is wired next to `VideoFilter`.

This is not #120 (hide-videos / MediaInfo). Not #98 (quoted video duration). Not #99 / #132 (self-reply chains).

## Five-line proof

- Entry: Following URT `QueryBuilder` (`server.rs`) and Night Owl `old_query` (`following_night_owl_source.rs`)
- Sink: `FollowingContentControlsFilter` on `ReverseChronPostsPipeline` (and Phoenix next to `VideoFilter`)
- Break: proto hide-replies / hide-links / exclude-replies / exclude-retweets were never copied; Night Owl hardcoded `include:retweets`
- Viewer effect: user who hid replies, links, or retweets still saw them on Latest Following
- Twin: `exclude_videos` is already copied and `VideoFilter` already exists (#120)

## Change

- Copy `hide_replies`, `hide_links`, `exclude_replies`, `exclude_retweets` onto `ScoredPostsQuery`
- Night Owl: `-filter:replies`, `-filter:links`, and `-filter:nativeretweets` when those flags are set
- Mixer filter drops replies / link cards (tweet, quoted, ancestor text) / retweets when the matching flag is on
- Wire the filter on reverse-chron and Phoenix

## Tests

- hide_replies / exclude_replies drop replies, keep originals
- hide_links drops https / t.co / quoted URL / `www.` host, keeps `notwww`
- exclude_retweets drops retweets, keeps replies
- Off flags keep every type
- Night Owl operators: default still `include:retweets`; flags compose
- Mock reverse-chron filter count includes the new filter

Standalone decision-table harness (same match arms): 19/19 passed.

## Leftover

For You cached-posts Redis key still keys only on `exclude_videos`. A cache hit can still contain replies/links; the new Phoenix filter drops them after the hit (slate may underfill). Same class as #98 leftover for quoted duration on cache.

Fork PR: none

### DIFF
--- a/home-mixer/candidate_pipeline/phoenix_candidate_pipeline.rs
+++ b/home-mixer/candidate_pipeline/phoenix_candidate_pipeline.rs
@@ -54,6 +54,7 @@ use crate::filters::retweet_deduplication_filter::RetweetDeduplicationFilter;
 use crate::filters::self_tweet_filter::SelfTweetFilter;
 use crate::filters::topic_ids_filter::TopicIdsFilter;
 use crate::filters::vf_filter::VFFilter;
+use crate::filters::following_content_controls_filter::FollowingContentControlsFilter;
 use crate::filters::video_filter::VideoFilter;
 use crate::filters::viewer_muted_keyword_filter::ViewerMutedKeywordFilter;
 use crate::models::candidate::PostCandidate;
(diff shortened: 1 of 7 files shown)


Pitchfork-and-Torch:

PR #140, opened 2026-09-07
Drop quotes whose quoted author handle matches a muted keyword

Pitchfork-and-Torch:
## Bug

FollowingViewerMutedKeywordFilter already drops when quoted tweet text matches a muted keyword. AuthorSocialgraphFilter already drops when the viewer mutes the quoted author by user id (PR 23). Mute-keyword matching never looked at the quoted author handle.

Gizmoduck already filled author_screen_name and retweeted_screen_name. get_screen_names() omitted quoted_user_id. Latest Following never fetched the quoted handle at all.

A followee quoting @spamaccount with comment "look" and quoted text "hello" still served when the viewer muted the keyword spamaccount. Mention-in-text already drops (test_mention_keyword). Quoted text already drops on Following. This is the leftover author-of-quoted edge.

This is not PR 96 (quote/reply/retweet TEXT). Not PR 23 (mute quoted author by id). Not PR 100 (blocked-by retweeted author). Not PR 108 (Following author blocked-by).

## Fix

Hydrate quoted_screen_name from Gizmoduck (Phoenix already had the client; Latest Following gets a quoted-user-only hydrator). Match that handle in FollowingViewerMutedKeywordFilter and ViewerMutedKeywordFilter.

## Proof

- Entry: Night Owl / QuoteHydrator quoted_user_id; Gizmoduck profile.screen_name
- Sink: FollowingViewerMutedKeywordFilter / ViewerMutedKeywordFilter
- Break: mute-keyword matched tweet_text and quoted_tweet_text; quoted author handle was never hydrated or matched
- Viewer effect: quote of @mutedkeyword still serves on Latest Following and For You
- Twin: test_mention_keyword drops @handle in tweet_text; Following already drops matching quoted_tweet_text

## Tests

- drops_quote_when_quoted_author_handle_matches_muted_keyword (Following and For You)
- keeps_quote_when_quoted_author_handle_does_not_match
- keeps_quote_when_only_quoted_text_would_match_keyword (For You does not clone PR 96 text matching)
- still_drops_when_quoted_text_matches (Following twin unchanged)
- unit tests added; cargo test cannot run in the public dump

PR #23, opened 2026-08-13
Fix mute/block asymmetry: drop quotes of muted authors

Pitchfork-and-Torch:
## Summary
AuthorSocialgraphFilter already drops a candidate when the viewer blocks the quoted author (quoted_user_id vs blocked_user_ids). Mute only checked candidate.author_id, so a quote of a muted user still served.

This PR clones the block twin for mute. One class, one file. No ranking changes. Retweet mute stays in PR 9.

## Proof
- Entry: muted_user_ids_query_hydrator.rs writes query.user_features.muted_user_ids
- Sink: AuthorSocialgraphFilter
- Break: mute matched author_id only; quoted_user_id was block-only
- Viewer effect: F quotes muted user Q; viewer still sees the quote
- Twin: viewer_blocks_quoted_author

## Test plan
- [ ] test_muted_quoted_author_is_removed fails on main, passes on this branch
- [ ] test_unmuted_quoted_author_is_kept still keeps a quote of an unmuted user
- [ ] unit tests added; cargo test cannot run in the public dump

PR #153, opened 2026-09-07
Stop WhoToFollow from recommending muted or blocked accounts

Pitchfork-and-Torch:
## Bug

WhoToFollow `excluded_user_ids` only sent previously shown WTF accounts. It never sent muted or blocked user ids.

For You and Latest Following blender pipelines never loaded those SGS lists. Phoenix loads them for post filters only. WTF runs at the blender layer, so VF never sees the recommended accounts.

SGS `get_muted_user_ids` / `get_blocked_user_ids` Err is ignored by query hydration (`update` is skipped). Empty lists look like "muted nobody".

This is not #137 (VF primary-author relationship Failed). This is not #23 / #140 (quoted-author mute on posts). ListTimeline / ListLatest are not in this dump.

## Five-line proof

- Entry: `MutedUserIdsQueryHydrator` / `BlockedUserIdsQueryHydrator` (SGS list membership). For You / Latest Following blender queries never ran them.
- Sink: `WhoToFollowSource::get_excluded_user_ids` → ARM `WhoToFollowReactiveContext.excluded_user_ids`
- Break: exclude set was fatigue-only; list Err left empty vectors; blender never hydrated the lists
- Viewer effect: a muted or blocked account can appear in Who To Follow on For You and Latest Following
- Twin: `FollowingRepliedUsersHydrator` already excludes muted+blocked ids; `AuthorSocialgraphFilter` already uses the same lists for posts

## Change

- Mark mute/block lists hydrated only after a successful SGS read. Err leaves the flags false.
- Load both lists on the For You and Latest Following blender pipelines.
- WhoToFollow stays off until both lists are ready. `excluded_user_ids` puts muted/blocked ids first, then fatigue ids.

Genuine empty lists (viewer muted/blocked nobody) still enable the module.

## Tests

- Enable is false until both list flags are set
- Muted and blocked ids are excluded before fatigue ids
- A full mute/block cap does not admit a fatigue id
- Hydrator `update` sets the ready flag

Standalone decision-table harness (same exclude / ready arms): 8/8 passed.

## Leftover

Mixer `AuthorSocialgraphFilter` still treats an unready empty list as nobody muted/blocked for quoted/RT authors. After #137, VF is the sink for primary-author mute/block. WTF has no VF sink.

Fork PR: none

### DIFF
--- a/home-mixer/candidate_pipeline/following_candidate_pipeline.rs
+++ b/home-mixer/candidate_pipeline/following_candidate_pipeline.rs
@@ -14,7 +14,9 @@ use crate::clients::who_to_follow_client::{
 use crate::filters::invalid_conversation_module_filter::InvalidConversationModuleFilter;
 use crate::models::query::ScoredPostsQuery;
 use crate::params;
+use crate::query_hydrators::blocked_user_ids_query_hydrator::BlockedUserIdsQueryHydrator;
 use crate::query_hydrators::followed_user_ids_query_hydrator::FollowedUserIdsQueryHydrator;
+use crate::query_hydrators::muted_user_ids_query_hydrator::MutedUserIdsQueryHydrator;
 use crate::query_hydrators::past_request_timestamps_query_hydrator::PastRequestTimestampsQueryHydrator;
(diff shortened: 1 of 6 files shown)


Pitchfork-and-Torch:

PR #118, opened 2026-09-06
Stop QuoteHydrator from skipping quoted ids on retweets

Pitchfork-and-Torch:
## Bug

Phoenix `QuoteHydrator` asks TES for quoted tweets by wrapper `tweet_id`.

`get_quoted_tweets` reads the quoting tweet's nested quote. That field lives on the original post. A retweet wrapper stores `source_tweet_id`, not `quoted_tweet`. `quoted_tweet_id` / `quoted_user_id` stay None.

OON retweets are already dropped by `OONRetweetReplyFilter`. This path is a followed-user retweet of a quote (in-network).

- Entry: Phoenix `QuoteHydrator` (`phoenix_candidate_pipeline.rs`, `phoenix_scores_pipeline.rs`)
- Sink: `VFCandidateHydrator` puts `quoted_tweet_id` on TimelineHomeRecommendations; `AuthorSocialgraphFilter` quoted mute/block; `AdsBrandSafetyVfHydrator` quoted labels
- Break: TES and cache keyed by `candidate.tweet_id` instead of `get_original_tweet_id()` (`retweeted_tweet_id.unwrap_or(tweet_id)`)
- Viewer effect: retweet of a quote of a VF-Drop / blocked / muted inner tweet still serves on For You
- Twin: `MediaInfoHydrator`, `LanguageCodeHydrator`, and `EngagementCountsHydrator` already cache and fetch by `get_original_tweet_id()`

This is not PR 117 (TES/graph fail-closed). This is not PR 94 (core TES wiping retweet/reply ids). This is not PR 98 (exclude_videos on quotes). This is not PR 23 (mute quoted author when the id is already present).

## Fix

Look up and cache TES quoted tweets by `candidate.get_original_tweet_id()`. Native quotes are unchanged (`original == tweet_id`). Two retweets of the same quote now share the cache key.

## Tests

- Native quote still hydrates when TES is keyed by the candidate id
- Retweet of a quote hydrates when TES only has the original id (fails on unmodified main)
- TES keyed only by the wrapper id does not hydrate a retweet of a quote

cargo: cannot run. Public dump has no Home Mixer manifest.

PR #96, opened 2026-09-04
Drop For You cards whose quote, reply, or retweet text matches muted keywords

Pitchfork-and-Torch:
## Bug
ViewerMutedKeywordFilter on Phoenix For You only tokenized candidate.tweet_text. Quotes, replies, and native retweets of muted-keyword posts still served. Native RTs often have empty tweet_text; the original lives on retweeted_tweet_id.

Following already matched quoted_tweet_text and ancestor_texts. It did not match retweet original text either.

This is not mute-on-quote author id (PR 23). Not mute-on-retweet author id (PR 9). Not VF dual-role merge (PR 55 / 90). Not conversation-gap VF ancestors (PR 97).

## Fix
Clone the Following matcher and add retweeted_tweet_text on For You and Following. Wire QuotedPostTextHydrator after QuoteHydrator on Phoenix. That hydrator fills quoted_tweet_text, ancestor_texts, and retweeted_tweet_text from TES. It does not expand the ancestor chain.

## Proof
- Entry: QuoteHydrator quoted_tweet_id; Thunder ancestors; TES retweeted_tweet_id
- Sink: ViewerMutedKeywordFilter / FollowingViewerMutedKeywordFilter
- Break: For You used tweet_text only; hydrator never fetched retweet original text
- Viewer effect: quote, reply, or RT of a muted-keyword post still serves
- Twin: Following quoted+ancestor matching; retweet original is the leftover both surfaces missed

## Tests
- drops_quote_whose_quoted_text_matches_muted_keyword
- drops_reply_when_ancestor_text_matches_muted_keyword
- drops_retweet_whose_original_text_matches_muted_keyword
- fills_retweeted_tweet_text
- fills_ancestor_texts_without_changing_ancestor_ids
- unit tests added; cargo test cannot run in the public dump

PR #155, opened 2026-09-07
Stop SubscriptionHydrator from skipping exclusive ids on retweets

Pitchfork-and-Torch:
Phoenix `SubscriptionHydrator` asks TES for exclusive authors by wrapper `tweet_id`. Super Follow lives on the original. A retweet wrapper is not exclusive, so `subscription_author_id` stays None and `IneligibleSubscriptionFilter` keeps the card.

- Entry: Phoenix `SubscriptionHydrator` (`phoenix_candidate_pipeline.rs`)
- Sink: `IneligibleSubscriptionFilter` keeps `subscription_author_id: None` as public
- Break: TES and cache keyed by `candidate.tweet_id` instead of `get_original_tweet_id()`
- Viewer effect: subscriber RT of an exclusive post still serves on For You to non-subscribers
- Twin: `MediaInfoHydrator`, `LanguageCodeHydrator`, `EngagementCountsHydrator` already fetch by original; QuoteHydrator is #118

This is not #118 (quoted ids). This is not #129 (subscription-list miss). This is not #141 (TES fail-closed).

Look up and cache TES exclusive authors by `candidate.get_original_tweet_id()`. Native exclusive posts are unchanged. Two retweets of the same exclusive now share the cache key.

- Native exclusive still keys TES by the candidate id
- Retweet of exclusive keys TES by the original id (fails on unmodified main)
- TES keyed only by the wrapper id does not treat the retweet as exclusive

Fork PR: none
Survey: #96–#142. Unclaimed.

### DIFF
--- a/home-mixer/candidate_hydrators/subscription_hydrator.rs
+++ b/home-mixer/candidate_hydrators/subscription_hydrator.rs
@@ -1,5 +1,5 @@
 use crate::clients::tweet_entity_service_client::TESClient;
-use crate::models::candidate::PostCandidate;
+use crate::models::candidate::{CandidateHelpers, PostCandidate};
 use crate::models::query::ScoredPostsQuery;
 use std::sync::Arc;
 use tonic::async_trait;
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #156, opened 2026-09-07
Rehydrate block and author-label bits on cached For You posts

Pitchfork-and-Torch:
## Bug

`BlockedByHydrator`, `GizmoduckCandidateHydrator`, and `QuoteHydrator` all use `enable = !has_cached_posts`.

`MutedUserIdsQueryHydrator` / `BlockedUserIdsQueryHydrator` still run on a cache hit. The lists are current. `AuthorSocialgraphFilter` can drop a newly muted or viewer-blocked author from those lists. It cannot recompute `author_blocks_viewer`, `quoted_author_blocks_viewer`, or `nsfw_author` — those bits were written up to 180s earlier.

`VFCandidateHydrator` is post-selection. `OONNsfwSimclustersFilter` reads `nsfw_author` before scoring. A newly labeled SimClusters author keeps `Some(false)` and stays in the slate.

This is not #135 (in_network bit). Not #125 (gizmoduck origin / miss). Not #137 (socialgraph RPC fail-open).

## Five-line proof

- Entry: `BlockedByHydrator` / `GizmoduckCandidateHydrator` / `QuoteHydrator` (`enable = !has_cached_posts`)
- Sink: `AuthorSocialgraphFilter` (`author_blocks_viewer` / `quoted_author_blocks_viewer`); `OONNsfwSimclustersFilter` (`nsfw_author == Some(true)`)
- Break: cache hit skipped the hydrators, so the Redis slate bits stayed
- Viewer effect: an author who newly blocked the viewer, a quote of one, or a newly labeled NSFW author still appears for up to 180s
- Twin: #135 already recomputes `in_network` from the live follow list on the same cache hit

## Fix

Always enable. `update` overwrites the cached bits.

On a cache hit, `QuoteHydrator` does not re-call TES. It keeps `quoted_tweet_id` / `quoted_user_id` already on the candidate and only refreshes `quoted_author_blocks_viewer`. A TES miss therefore cannot wipe the quoted author id.

## Tests

- `enable` is true on cache hit and miss (all three hydrators)
- cache-hit BlockedBy: stale `Some(true)` → `false`, newly blocking → `true`
- cache-hit Quote: quoted ids and duration kept; newly blocking quoted author → `true`; previously blocking → `false`

Standalone decision-table harness (same enable + filter predicates): 14 assertions passed.

Fork PR: none

### DIFF
--- a/home-mixer/candidate_hydrators/blocked_by_hydrator.rs
+++ b/home-mixer/candidate_hydrators/blocked_by_hydrator.rs
@@ -17,8 +17,11 @@ impl BlockedByHydrator {
 
 #[async_trait]
 impl Hydrator<ScoredPostsQuery, PostCandidate> for BlockedByHydrator {
-    fn enable(&self, query: &ScoredPostsQuery) -> bool {
-        !query.has_cached_posts
+    fn enable(&self, _query: &ScoredPostsQuery) -> bool {
+        // MutedUserIds / BlockedUserIds query hydrators still run on cache
+        // hits. Skipping here keeps author_blocks_viewer from the Redis slate
+        // (TTL 180s). A newly blocking author then still serves.
+        true
     }
 
     async fn hydrate(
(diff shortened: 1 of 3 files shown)


Pitchfork-and-Torch:

Issue #41, opened 2026-08-14
Why isn’t my post reaching her timeline? 😭


ig-imanish:
**Description:**  
My post is not reaching her timeline.  

Not sure if it’s a visibility issue, algorithm problem, or something else. Any idea why this is happening?

jeheskielSunloy77:
appearance issue maybe bro? (speaking from experience)

tomihq:
bro, this issue 😭😭

FinnT730:
maybe because you are not what she likes, so your content does not need to reach her.
I have no idea who "her" even is...
but if they are not following you, then tough luck

nvhai245:
this is in fact, a skill issue.

PR #157, opened 2026-09-07
Stop Explore from ranking posts when topic hydration is missing

Pitchfork-and-Torch:
## Problem

Bulk Explore (`topic_ids.len() > 6`) is an exclusion filter. A missing topic hydration (`filtered_topic_ids = None`) was treated as keep. `FilteredTopicsHydrator` also wrote `None` on a successful empty lookup, so a Strato miss and a real untagged post were the same value.

TopicIdsFilter runs before VF. Withheld Sports/Politics posts whose topic row failed to load still ranked on Explore.

This is not #122 (Phoenix request-side bitmap). Same class, later gate.

## Change

- `None` = fetch/decode miss → drop on bulk Explore (fail closed).
- `Some([])` = successful lookup, no topics → still allowed as untagged filler.
- Successful hydrator decodes now write `Some([])` so the two cases stay distinct.

## Proof

1. `is_bulk_topic_request` is `topic_ids.len() > 6` (Explore multi-topic page).
2. Hydrator `StratoResult::Err` / empty bytes left `filtered_topic_ids = None`.
3. Bulk match arm `_ => true` kept that candidate.
4. TopicIdsFilter is pre-rank; VF is post-selection — withheld topics occupied Explore slots.
5. After: `None => false`, `Some([]) => true`; tests `test_bulk_explore_drops_hydration_miss_not_known_untagged` and updated `test_bulk_filter_removes_only_excluded_topics`.

## Survey

- Pitchfork-and-Torch/x-algorithm open PRs: **none** (fork #41 opened by tool default and closed immediately).
- Fork PR: **none**.
- Distinct from upstream #122 (retrieval bitmap on Sports/Science/Technology).

Base: `xai-org/x-algorithm` `main`. Head: `Pitchfork-and-Torch/cursor/explore-topic-hydration-fail-open-9d2a`.

### DIFF
--- a/home-mixer/candidate_hydrators/filtered_topics_hydrator.rs
+++ b/home-mixer/candidate_hydrators/filtered_topics_hydrator.rs
@@ -22,14 +22,24 @@ fn decode_topics_pair(
             match decoded {
                 StratoResult::Ok(v) => {
                     let ft = v.v;
-                    let exp_topics = ft
-                        .as_ref()
-                        .and_then(|ft| ft.topic_ids_for_experiment(experiment).cloned());
+                    // Some([]) = lookup succeeded with no topics. None is
+                    // reserved for fetch/decode failure so Explore can fail closed.
+                    let exp_topics = Some(
+                        ft.as_ref()
+                            .and_then(|ft| ft.topic_ids_for_experiment(experiment).cloned())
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:

PR #161, opened 2026-09-07
Fail closed when TES media or URL entity lookup is missing

Pitchfork-and-Torch:
## Bug

TES `get_tweet_media_entities` `Ok(None)` / Failed assembled as `has_media=false` via `get_or_default`. `nsfw_base_condition` and `NsfwAuthorInterstitialRule` require `has_media()`, so NSFW_HIGH_PRECISION age-gating and author interstitial skipped. `get_urls` was never called, so a card URL RPC miss was invisible.

This is not #104 (bot 7429 cleanup). This is not #107 (GetTweetLabelInfoFromURL RTF reads). This is not #139 (SafetyLabelHydrator empty map). This is not #134 (TES flag Failed only; that leftover left media NotFound as no-media).

Found(empty) is a confirmed no-media / no-card tweet. That path is unchanged.

## Five-line proof

- Entry: `TesHydrator::hydrate_tweets` (`get_tweet_media_entities`, now also `get_urls`)
- Sink: `retain_candidates_with_usable_media_or_url_entities` omit → `FilterTweets` `Verdict::unresolved_author` Drop
- Break: media NotFound/Failed became has_media=false; URL entities were never fetched
- Viewer effect: NSFW_HIGH_PRECISION / NSFW-author media or card whose TES entity missed still served
- Twin: #134 fail-closes media Failed only and leaves NotFound; #139 is the label store

## Change

Drop a candidate when the media slot is NotFound, Failed, or missing. Fetch URL entities; drop on URL Failed. Confirmed empty media/URL lists still assemble and serve.

## Tests

- Found empty media + urls is usable
- Media NotFound / Failed is unusable
- URL NotFound is usable; URL Failed is not
- Retain drops only unusable ids

Standalone decision-table harness (same match arms): 12/12 passed.

`cargo test` cannot run. Public dump has no visibility-filtering crate manifest.

Fork PR: none

### DIFF
--- a/visibility-filtering/hydration/mod.rs
+++ b/visibility-filtering/hydration/mod.rs
@@ -25,7 +25,7 @@ use safety_label_hydrator::{SafetyLabelHydration, SafetyLabelHydrator};
 use socialgraph_hydrator::SocialgraphHydrator;
 use std::collections::HashMap;
 use std::sync::Arc;
-use tes_hydrator::TesHydrator;
+use tes_hydrator::{retain_candidates_with_usable_media_or_url_entities, TesHydrator};
 use viewer_hydrator::ViewerHydrator;
 use xai_core_entities::gizmoduck_client::GizmoduckClient;
 use xai_core_entities::tweet_entity_service_client::TESClient;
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:

PR #162, opened 2026-09-07
Fail closed when Flock mute/block graph slots are missing

Pitchfork-and-Torch:
## Bug

`decode_relationship_edges` mapped a truncated Flock multi-query onto empty mute/block/mute-retweet sets. RPC success with 1–3 result slots therefore assembled `viewer_mutes_author = false` / `viewer_blocks_author = false`.

A total Flock `Err` already returns an empty map (`SocialgraphHydrator` marks Failed). Four empty slots is a confirmed unlabeled read. Those paths are unchanged.

This is not #137 (total Failed assembled via `get_or_default`). #137 leftover: "Partial Flock result slots still default missing mute/block graphs to empty."

## Five-line proof

- Entry: `decode_relationship_edges` (`next_edge_set` used `unwrap_or_default` for missing slots)
- Sink: `ViewerMutesAuthorRule` / `ViewerBlocksAuthorRule` / `MutedRetweetsRule`
- Break: 1-slot Flock success decoded as follows-only, mute/block empty → Allow
- Viewer effect: a muted or blocked author still served when Flock returned a truncated multi-query
- Twin: #137 fail-closes a total Failed read; four confirmed-empty slots stay unlabeled

## Change

Require exactly four result slots (follows, blocks, mutes, mute-retweets). Otherwise return no relationships, same sink as RPC `Err`. Four empty slots remain a successful all-false read.

## Tests

- Four named slots still decode follows/blocks/mutes/mute-retweets
- Four empty slots stay confirmed unlabeled
- 0 / 1 / 3 / 5 slots fail closed (`None`)

Standalone decision-table harness (same match arms): 8 assertions passed.

`cargo test` cannot run. Public dump has no visibility-filtering manifest.

Fork PR: none

### DIFF
--- a/visibility-filtering/clients/socialgraph_client.rs
+++ b/visibility-filtering/clients/socialgraph_client.rs
@@ -111,6 +111,8 @@ fn relationship_select_request(viewer_id: u64, destination_ids: &[i64]) -> Selec
     }
 }
 
+const RELATIONSHIP_GRAPH_SLOTS: usize = 4;
+
 fn next_edge_set(results: &mut impl Iterator<Item = Results>) -> HashSet<u64> {
     results
         .next()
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #175, opened 2026-09-08
Recompute mutual-follow bits on cached For You posts

Pitchfork-and-Torch:
## Bug

`BidirectionalFollowHydrator.enable` was `flag && !has_cached_posts`.

`FollowedUserIdsQueryHydrator` still runs on a cache hit. The follow list is current. `RankingScorer` still runs. It reads `is_mutual_follow_author` and, when that bit is `Some(true)`, adds `BidirectionalFollowReplyWeightBoost` to the reply head. That boost defaults to 15.0 on a `ReplyWeight` of 5.0.

Skipping the hydrator keeps the bit written up to 180s earlier. A new mutual does not get the boost. A broken mutual keeps a 4x reply-head weight. `PhoenixScorer` is also skipped on the cache hit, so the weighted score is rebuilt from cached phoenix heads times the stale bit.

This is not #135 (`in_network` bit). Not #156 (block / nsfw_author / quote).

## Five-line proof

- Entry: `BidirectionalFollowHydrator` (`enable` skipped `has_cached_posts`)
- Sink: `RankingScorer::reply_weight_for` / `dwell_weight_for` (`is_mutual_follow_author == Some(true)`)
- Break: cache hit left the Redis slate bit in place while the follow list was live
- Viewer effect: originals from a new mutual stay unboosted; originals from a broken mutual keep a 4x reply-head boost for up to 180s
- Twin: #135 recomputes `in_network` on the same cache hit; #156 recomputes block / nsfw bits

## Fix

Enable whenever the hydration flag is on. `update` overwrites the cached bits.

## Tests

- `enable` is true on cache hit and miss
- cache-hit hydrate + update: author no longer follows viewer -> `Some(false)`
- cache-hit hydrate + update: viewer unfollowed -> `Some(false)`
- cache-hit hydrate + update: newly mutual -> `Some(true)`

`RankingScorer` test `bidirectional_weight_boosts_only_mutual_original_posts` is the score sink.

### DIFF
--- a/home-mixer/candidate_hydrators/bidirectional_follow_hydrator.rs
+++ b/home-mixer/candidate_hydrators/bidirectional_follow_hydrator.rs
@@ -14,7 +14,7 @@ pub struct BidirectionalFollowHydrator {
 #[async_trait]
 impl Hydrator<ScoredPostsQuery, PostCandidate> for BidirectionalFollowHydrator {
     fn enable(&self, query: &ScoredPostsQuery) -> bool {
-        query.params.get(EnableBidirectionalFollowHydration) && !query.has_cached_posts
+        query.params.get(EnableBidirectionalFollowHydration)
     }
 
     async fn hydrate(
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #176, opened 2026-09-08
Rescore cached For You posts when the action sequence is newer

Pitchfork-and-Torch:
Cache hits still hydrate a fresh UAS scoring sequence, then PhoenixScorer disabled itself whenever has_cached_posts was set. RankingScorer kept blending the Redis-frozen Phoenix heads (VQV, dwell, fav) for the full 180s TTL, so For You order ignored videos the viewer just finished or skipped.

Proof
- Entry: ScoringSequenceQueryHydrator writes query.scoring_sequence on every request, including cache hits.
- Sink: PhoenixScorer.enable / RankingScorer.compute_weighted_parts (vqv_score, dwell, fav) into For You score.
- Break: enable() returned false on has_cached_posts, so last_scored_at_ms heads older than metadata.last_modified_epoch_ms / last_sequence_time never refreshed.
- Viewer: after a VQV or skip, pull-to-refresh still ranks the same cached slate as if those actions never happened.
- Twin: metadata.length already gates new-user cluster selection in this same scorer; cache skip is the only path that ignores a newer sequence.

Fix: keep the Redis candidate cache, but run Phoenix again when any cached last_scored_at_ms is missing or older than the hydrated sequence. Unchanged cache hits stay skipped. Missing sequence stays skipped so update() cannot wipe cached heads with Default scores.

### DIFF
--- a/home-mixer/scorers/phoenix_scorer.rs
+++ b/home-mixer/scorers/phoenix_scorer.rs
@@ -58,19 +58,54 @@ impl PhoenixScorer {
 
         configured_cluster
     }
+
+    fn ranker_killed(query: &ScoredPostsQuery) -> bool {
+        query
+            .decider
+            .as_ref()
+            .is_some_and(|d| d.enabled(PHOENIX_RANKER_KILL_SWITCH_DECIDER))
+    }
+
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #178, opened 2026-09-08
Stop ancillary VF from dropping followee quotes of civic-labeled posts

Pitchfork-and-Torch:
## Bug

For You fetches quoted posts and conversation ancestors at TimelineHomeRecommendations. `should_drop_ancillary` treats any Recs Drop on those ids as a drop of the card.

`FosnrCivicIntegrityDropRule` is in the shared Home table, so Recs Drops civic-labeled posts. A followee quote or reply then dies in `AncillaryVFFilter` even when the followee's own post has no civic label.

Civic integrity is a do-not-amplify mark on the attached post. A quote or reply is the followee's speech. Hate / abuse / violent-speech ancillary drops are unchanged. DoNotAmplify tweet-label drops on quotes are unchanged.

This is not the civic Home-table move. This is not retweet source-label hydration. This is not Following quotes scored at Home.

## Fix

Emit civic drops as `SafetyResult` / `FosnrCivicIntegrity` so the mixer can tell them apart from generic `PossiblyUndesirable` (spam, DNA, abuse-insults).

In-network quotes and ancestor replies skip that civic Drop. Out-of-network quotes still drop. Retweets of a civic-labeled original still drop.

The labeled civic post itself still Drops on Home and Recs.

## Proof

- Entry: `VFCandidateHydrator` puts `quoted_tweet_id` and `ancestors` on TimelineHomeRecommendations
- Sink: `should_drop_ancillary` -> `AncillaryVFFilter`
- Break: civic Recs Drop on the attached id removed the followee card
- Viewer effect: followed quote or reply of a civic-labeled post vanishes from For You
- Twin: `DoNotAmplifyOonDropRule` still drops a quote of a DNA-labeled post

## Tests

- In-network quote of civic: keep
- In-network reply with civic ancestor: keep
- OON quote of civic: drop
- In-network retweet of civic: drop
- In-network quote of DoNotAmplify: still drop
- Missing `in_network` on a civic quote: drop
- Golden: civic still Drops on Home and Recs via `FosnrCivicIntegrityDropRule`

`cargo test` cannot run. Public dump has no Home Mixer or VF crate manifest. Mixer match logic was compiled and run as a rustc model: 6/6 passed.

### DIFF
--- a/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
+++ b/home-mixer/candidate_hydrators/vf_candidate_hydrator.rs
@@ -9,7 +9,7 @@ use tonic::async_trait;
 use xai_candidate_pipeline::hydrator::Hydrator;
 use xai_twittercontext_proto::GetTwitterContextViewer;
 use xai_twittercontext_proto::TwitterContextViewer;
-use xai_visibility_filtering::models::{Action, FilteredReason};
+use xai_visibility_filtering::models::{Action, FilteredReason, SafetyResultReason};
 use xai_visibility_filtering::vf_client::SafetyLevel;
 use xai_visibility_filtering::vf_client::SafetyLevel::{TimelineHome, TimelineHomeRecommendations};
 use xai_visibility_filtering::vf_client::{TweetVisibility, VfClient};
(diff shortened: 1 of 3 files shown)


Pitchfork-and-Torch:

PR #179, opened 2026-09-08
Stop seen/served filters from dropping replies to seen parents

Pitchfork-and-Torch:
## Bug

`related_post_ids_iter` chained `in_reply_to_tweet_id` next to the candidate id and the retweet original. PreviouslySeen, PreviouslySeenBackup, and PreviouslyServed hard-drop a candidate when any related id is in seen / impressed / served.

A reply is new speech. Seeing or serving the parent is not seeing the reply. Followee replies to a post the viewer already scrolled past never reach For You.

This is not #151 (quote same-card). #151 adds `quoted_tweet_id`. This removes the parent id. A retweet of a seen original still drops.

## Five-line proof

- Entry: TES / Thunder `in_reply_to_tweet_id` on an in-network reply
- Sink: PreviouslySeen / PreviouslySeenBackup / PreviouslyServed (`related_post_ids_iter`)
- Break: parent id was treated as the reply card
- Viewer effect: a followee reply to a seen/served/impressed post disappears from For You
- Twin: same reply kept when the parent is absent from those sets; RT of a seen original still drops

## Fix

Same-card ids are the post and its retweet original only.

## Tests

- related_ids_are_the_card_and_retweet_original_only
- keeps_reply_when_only_the_parent_was_seen
- drops_reply_when_the_reply_itself_was_seen
- keeps_reply_when_only_parent_was_impressed
- keeps_reply_when_only_the_parent_was_served

Fork PR: none

### DIFF
--- a/home-mixer/filters/previously_seen_posts_backup_filter.rs
+++ b/home-mixer/filters/previously_seen_posts_backup_filter.rs
@@ -78,6 +78,23 @@ mod tests {
     }
 
     #[test]
+    fn test_keeps_reply_when_only_parent_was_impressed() {
+        let query = ScoredPostsQuery {
+            impressed_post_ids: vec![10],
+            ..Default::default()
+        };
+        let reply = PostCandidate {
+            tweet_id: 20,
+            in_reply_to_tweet_id: Some(10),
+            ..Default::default()
+        };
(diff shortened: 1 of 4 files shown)


Pitchfork-and-Torch:

PR #173, opened 2026-09-08
Stop safemodel R/X scores from applying hard-NSFW drop labels

Pitchfork-and-Torch:
## Bug

The PTOS write sink treated safemodel sex-and-nudity R/X scores as AdultContentSexualHard when the policy model found no hard-NSFW hit.

Safemodel is a high-recall media classifier (`R` is not a policy violation). Deluxe runs it on every media post. The sink then applied NsfwHighRecall and NsfwHighPrecision, which VF drops on TimelineHomeRecommendations. News, art, protest, and other non-violating speech lost OON For You reach.

PTOS AdultContentSexualHard still applies those labels. Comparison metrics stay.

## Proof

- Entry: grox PlanSafetyPtos task_safety_ptos_safemodel_sex_nudity (R/X buckets; deluxe has no adult-suspicion gate)
- Sink: TaskWriteSafetyPostAnnotationsResultSink then VF NsfwHighRecallDropRule / NsfwHighPrecisionOonDropRule
- Break: `if safemodel.positive and not ptos_already_nsfw` minted SexualHard and applied drop labels
- Viewer effect: non-violating OON post disappears from For You; followers still see it on TimelineHome
- Twin: safemodel compare outcome `safemodel_only_positive` is shadow, not enforce

## Tests

- test_safemodel_only_positive_does_not_apply_drop_labels
- test_ptos_hard_nsfw_already_labeled_does_not_need_safemodel
- python3 -m unittest grox.flows.ptos.test_safemodel_enforce

PR #172, opened 2026-09-08
Stop RankingScorer author diversity from keying the retweeter

Pitchfork-and-Torch:
## Bug

`RankingScorer::author_pool_counts` decayed on wrapper `author_id`. Phoenix ranking features already key the original author (`as_tweet_info` / `get_original_author_id`). Distinct retweeters of one original did not share a diversity bucket. One retweeter of distinct originals did.

This is not #31 (author-size IPS). This is not #125 (gizmoduck origin size/NSFW). This is not #128/#154/#135 (in_network stamps). `in_network` stays the poster. Diversity is the post-Phoenix author-repeat lever.

## Five-line proof

- Entry: `RankingScorer::author_pool_counts` (`EnableAuthorDiversity`)
- Sink: `PhoenixCandidatePipeline` scorers `[PhoenixScorer, RankingScorer, VMRanker]` then `TopKScoreSelector`
- Break: pool count used retweeter `author_id` while Phoenix hashed `get_original_author_id()`
- Viewer effect: one original could take many For You slots via different retweeters; a single retweeter of distinct authors was starved
- Twin: `PostCandidate::as_tweet_info` / `build_tweet_infos` already send the original author to Phoenix

## Change

- Count author diversity on `get_original_author_id()` (retweeted user, else poster)
- Originals and quotes unchanged (no `retweeted_user_id`)

## Tests

- Two retweets of the same original author from different retweeters: second is decayed
- One retweeter of two original authors: no decay
- Existing same-poster originals still decay

Standalone rustc model of `author_pool_counts` + `get_original_author_id`: 3/3 passed.

`cargo test` cannot run. Public dump has no home-mixer crate manifest.

Fork PR: none

PR #181, opened 2026-09-08
Stop engagement-count store errors from zeroing For You rank features

Pitchfork-and-Torch:
Dragonfly Err in EngagementCountsHydrator was coerced to an empty HashMap. Every candidate then got Ok(CachedCounts::default()). CachedHydrator caches Ok for 60s. as_tweet_info unwrap_or(0) stamps Phoenix InputBuffer and Thompson sampling as if the post had no favs, replies, or views.

Proof
- Entry: EngagementCountsHydrator::hydrate_from_client (get_engagement_counts)
- Sink: PostCandidate::as_tweet_info -> Phoenix stamp_engagement_counts; AuthorColdStart Thompson (view_count_on_home, fav_count)
- Break: unwrap_or_else empty map -> Ok(default) -> Moka insert
- Viewer: a viral post ranks like a zero-engagement post for up to 60s after a store blip; the whole slate can collapse together
- Twin: LanguageCodeHydrator and MediaInfoHydrator return Err on store error, so update_all skips and the cache does not store empties

This is not #171/#177 (notes). Not #172 (author diversity retweeter key). Not #173 (safemodel). Not #174 (UserCred). Not #175/#176 (cache mutual-follow / Phoenix rescore). Not #142 (empty Phoenix heads). Not #22 (cold-start thresholds). Not #21 (DPP missing embeddings).

Fix: on client Err, return Err for candidates that need a fetch. CachedHydrator does not insert. update_all leaves existing counts. Successful per-id misses stay Ok(default), same as a tweet with no row.

Tests cover store Err (no zero stamp), no cache poison, cached-post ineligible preserve, and successful miss unchanged.

cargo test cannot run here. Public dump has no Home Mixer manifest. Standalone rustc model of the stamp path passed 4/4.

### DIFF
--- a/home-mixer/candidate_hydrators/engagement_counts_hydrator.rs
+++ b/home-mixer/candidate_hydrators/engagement_counts_hydrator.rs
@@ -131,16 +131,30 @@ impl CachedHydrator<ScoredPostsQuery, PostCandidate> for EngagementCountsHydrato
         unique_ids.sort_unstable();
         unique_ids.dedup();
 
+        // Store errors must not become Ok(empty). CachedHydrator caches Ok, and
+        // as_tweet_info unwrap_or(0) would stamp Phoenix / Thompson as zero
+        // engagement for 60s. LanguageCode / MediaInfo already return Err here.
         let counts = if unique_ids.is_empty() {
-            HashMap::new()
+            Ok(HashMap::new())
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #184, opened 2026-09-08
Stop ads served history from burying organic For You posts

Pitchfork-and-Torch:
## Bug

Ads bleed into organic For You via served history.

`UpdateServedHistorySideEffect` writes ads as `PROMOTED_TWEET` with `tweet_id = ad.post_id`. `recently_served_ids` ignored `entity_type` and copied every `tweet_id` / `source_tweet_id` into `query.served_ids`. `PreviouslyServedPostsFilter` then drops any organic candidate whose id is in that set.

Two effects:

1. One ad impression buries the organic post for the exclude window (default 10 minutes, 100 ids).
2. Ad rows consume the `ExcludeServedTweetIdsNumber` budget, so recent organic served ids fall off and can repeat.

Not brand-safety rank leak (#127). Not seen/served reply-parent (#179). Not quoted related ids (#151).

## Proof

```
INPUT:  ad.post_id=P served as PROMOTED_TWEET; organic candidate P on next For You
PATH:   put(ad) -> MH key stamps served_time_ms
        -> recently_served_ids copies P (no entity_type gate)
        -> PreviouslyServedPostsFilter.related_post_ids contains P
OUTPUT: organic P DROPPED
        ads also take max_items slots so older organic served ids fall off
```

## Fix

Allowlist `EntityIdType::TWEET` in `recently_served_ids`. Ads stay in served history for ad fatigue. Organic exclude-already-served stays organic.

Tests: promoted ids stay out; ads do not eat the budget; organic TWEET / retweet source ids still exclude; PreviouslyServedPostsFilter keeps P after an ad impression and still drops P after an organic impression.

### DIFF
--- a/home-mixer/query_hydrators/served_history_query_hydrator.rs
+++ b/home-mixer/query_hydrators/served_history_query_hydrator.rs
@@ -81,17 +81,21 @@ fn recently_served_ids(
 ) -> Vec<u64> {
     let min_time_ms = now_ms - (duration_minutes as i64 * 60_000);
 
+    // Organic exclude-already-served is TWEET-only. Ads are written as
+    // PROMOTED_TWEET with tweet_id = ad.post_id; folding those ids into
+    // served_ids lets one ad impression bury the organic post and also
+    // consume the max_items budget so recent organic ids fall off.
     history
         .iter()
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

PR #182, opened 2026-09-08
Drop For You posts whose author handle matches a muted keyword

Pitchfork-and-Torch:
## Bug
Gizmoduck already writes `author_screen_name` on For You. `ViewerMutedKeywordFilter` only matched `tweet_text`.

A muted keyword that is the card author's handle never hid the post when the body was clean. `@mutedkeyword` posting `"hello"` still ranked.

This is not quoted-author handle mute-keyword (#140). Not RT original-author handle mute-keyword (#164). Not Phoenix reply-parent VF (#180). Not notes x ranking (#171 / #177).

## Fix
Tokenize the already-hydrated `author_screen_name` with the same mute-keyword matcher as tweet text.

## Proof
- Entry: GizmoduckCandidateHydrator writes `author_screen_name` before ViewerMutedKeywordFilter
- Sink: ViewerMutedKeywordFilter on PhoenixCandidatePipeline
- Break: filter walked `tweet_text` only; card handle was never tokenized
- Viewer effect: @mutedkeyword posts "hello" still serve on For You
- Twin: test_mention_keyword already drops @handle in tweet text; #140 / #164 cover quote and RT handles only

## Tests
- drops_when_author_handle_matches_muted_keyword
- keeps_when_author_handle_does_not_match
- unit tests added; cargo test cannot run in the public dump (no home-mixer Cargo.toml)

PR #108, opened 2026-09-04
Drop Following posts whose author blocked the viewer

Pitchfork-and-Torch:
## Bug
FollowingBlockedByHydrator asked socialgraph whether quoted_user_id or retweeted_user_id blocked the viewer. It never sent candidate.author_id. AuthorSocialgraphFilter then drops on author_blocks_viewer (unwrap_or false).

A native Following post from an author who blocked the viewer still served: author_blocks_viewer was written from retweeted_user_id only, so a non-retweet was Some(false). VFFilter None is fail-open, but this path is worse: it claims the author did not block the viewer.

Phoenix BlockedByHydrator already checks author_id. PR 100 added retweeted_user_id on For You. This is the Following twin of that missing native-author edge. This is not mute-on-retweet (PR 9) and not quote mute (PR 23).

## Fix
Include author_id in check_blocked_by. Set author_blocks_viewer if either the primary author or the original retweeted author blocked the viewer. Keep quoted_author_blocks_viewer on quoted_user_id.

## Proof
- Entry: FollowingBlockedByHydrator check_blocked_by on reverse-chron Following
- Sink: AuthorSocialgraphFilter author_blocks_viewer
- Break: Following never passed author_id into blocked-by
- Viewer effect: user blocked by A still sees A's native Following post (and A's retweet of B)
- Twin: Phoenix BlockedByHydrator author_id (PR 100 is the retweet original twin)

## Tests
- native_author_who_blocked_viewer_is_marked (fails on unmodified main)
- retweet_of_author_who_blocked_viewer_is_still_marked
- retweeter_who_blocked_viewer_is_marked
- quoted_author_who_blocked_viewer_is_still_marked
- unit tests added; cargo test cannot run in the public dump (no Home Mixer Cargo.toml)

PR #186, opened 2026-09-08
Stop Thunder from burying undeleted in-network posts

Pitchfork-and-Torch:
## Bug

Thunder delete has a handler. Thunder undelete does not. A restored followee tweet stays off For You until the tombstone ages out.

## Five-line proof

1. `TweetDeleteEvent` removes the CompactPost and inserts `post_id` into `deleted_posts`.
2. Live Thrift union includes `TweetUndeleteEvent` with the same tweet/user payload as create.
3. Feeder match fell through to `_ => {}`, so no `InNetworkEvent` was emitted.
4. Serving `insert_posts` refused any later create for a tombstoned id (`if deleted_posts.contains_key { continue }`).
5. `GetInNetworkPosts` therefore omitted the restored tweet for up to the retention window.

Twin: delete is handled; undelete is not. This is not #138 (QuotedTweetDelete kills the quote) and not #150 (RT of a tombstoned original).

## Fix

- Feeder maps `TweetUndeleteEvent` through the same LightPost builder as create (nullcast still skipped) and emits `TweetCreateEvent`.
- Store clears the tombstone on re-insert. If the author deque still holds the ghost id, do not push a second index row.

## Tests

- `undelete_clears_tombstone_and_restores_post` (fails on unmodified main)
- Undelete maps to the same LightPost as create
- Nullcast undelete is skipped like create

cargo: cannot run. Public dump has no Thunder manifest.

## Rejection table (this wave)

| Candidate | Why not |
|---|---|
| OON followee RT via stale in_network | Same class as #128/#154/#183 |
| VF author_blocks_viewer missing | Live For You uses BlockedByHydrator; Following leak is #108 |
| RT originals scored at TimelineHome | #116 closed; #119 covers original verdicts |
| Author handle mute keyword | #182 claimed |
| Topic snooze untagged | #170 claimed |
| SidTail / PopularTopics / BroadcastLiveness enable | Forbidden intern-enable |
| Author-size IPS | Forbidden |
| Home-mixer TES/gizmoduck/VF fail-closed spray | Forbidden |

### DIFF
--- a/thunder/kafka/tweet_events_listener.rs
+++ b/thunder/kafka/tweet_events_listener.rs
@@ -73,6 +73,30 @@ fn is_eligible_video(tweet: &Tweet) -> bool {
         .unwrap_or(false)
 }
 
+fn light_post_from_event_tweet(
+    tweet: Option<&Tweet>,
+    user: Option<&crate::schema::user::User>,
+) -> Option<LightPost> {
+    let tweet = tweet?;
+    let core_data = tweet.core_data.as_ref()?;
+    if core_data.nullcast.unwrap_or(false) {
+        return None;
+    }
+    Some(LightPost {
+        post_id: tweet.id?,
+        author_id: user?.id?,
+        created_at: core_data.created_at_secs?,
+        in_reply_to_post_id: core_data.reply.as_ref().and_then(|r| r.in_reply_to_status_id),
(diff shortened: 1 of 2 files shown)


Pitchfork-and-Torch:

PR #180, opened 2026-09-08
Stop For You VF from skipping Phoenix reply parents

Pitchfork-and-Torch:
## Bug
Phoenix retrieval, MOE, Topics, and TweetMixer set `in_reply_to_tweet_id` and leave `ancestors` empty. Thunder and Latest Following copy the parent into `ancestors`.

`VFCandidateHydrator` only fetches `candidate.ancestors` as ancillary ids. `should_drop_ancillary` only walks that list. A mute or block on the parent tweet never reaches `AncillaryVFFilter`, so the reply still ranks on For You.

This is not conversation-gap expansion (#97). Not mixer `ancestor_users` mute/block (#101 / #111). Not quote/RT handle mute-keyword (#140 / #164). Not civic followee quotes (#178). Not seen/served reply parents (#179).

## Fix
When `ancestors` is empty, treat `in_reply_to_tweet_id` as the reply parent for VF fetch and ancillary drop. Thunder's existing ancestor list is unchanged.

## Proof
- Entry: PhoenixSource / PhoenixMOESource / PhoenixTopicsSource / TweetMixerSource write `in_reply_to_tweet_id` and leave `ancestors` empty
- Sink: VFCandidateHydrator oon_ids + should_drop_ancillary + AncillaryVFFilter
- Break: ancillary VF walked `ancestors` only; OON Phoenix replies never populated that list
- Viewer effect: reply to a muted or blocked parent still serves on For You
- Twin: ThunderSource and Following Night Owl already push `in_reply_to` into `ancestors`

## Tests
- reply_ancestor_ids_uses_in_reply_to_when_ancestors_empty
- reply_ancestor_ids_keeps_thunder_ancestors
- phoenix_reply_drops_when_parent_author_is_muted
- phoenix_reply_drops_when_parent_author_blocked_viewer
- phoenix_reply_keeps_when_parent_is_allowed
- thunder_ancestors_still_drop_and_do_not_use_a_stale_in_reply_to
- unit tests added; cargo test cannot run in the public dump (no home-mixer Cargo.toml)

PR #101, opened 2026-09-04
Drop replies whose conversation ancestors are muted or blocked

Pitchfork-and-Torch:
## Summary
AuthorSocialgraphFilter already drops when the viewer mutes or blocks the candidate author, and when the viewer blocks a quoted or retweeted author. Conversation modules still render ancestor tweets (root plus parent) from ancestor_users. Mute and block did not look at that list.

Brazil2026ElectionFilter already drops a reply when ancestor_users hits the listed-author set, with an explicit comment that ancestors are returned on the scored post for For You thread UI. This PR clones that ancestor edge for muted_user_ids and blocked_user_ids. One class, one file. No ranking changes.

This is not mute-on-quote (PR 23), not mute-on-retweet (PR 9), and not retweet original blocked-by (PR 100).

## Proof
- Entry: CoreDataCandidateHydrator.build_ancestor_users writes candidate.ancestor_users (parent author plus root author)
- Sink: AuthorSocialgraphFilter (Phoenix pre-selection and Following reverse-chron post-selection)
- Break: mute/block matched author_id, quoted_user_id, and retweeted_user_id; ancestor_users was Brazil-only
- Viewer effect: F replies to muted or blocked user Q; Home still serves the reply and convo_marshaller includes Q's tweet in the conversation module (unless tombstoned)
- Twin: Brazil2026ElectionFilter.should_remove ancestor_users any(is_excluded); FollowingViewerMutedKeywordFilter already matches ancestor_texts

## Test plan
- [ ] test_muted_ancestor_author_is_removed fails on main, passes on this branch
- [ ] test_blocked_ancestor_author_is_removed fails on main, passes on this branch
- [ ] test_unmuted_unblocked_ancestor_is_kept still keeps a reply whose ancestors are not muted or blocked
- [ ] unit tests added; cargo test cannot run in the public dump (no Home Mixer Cargo.toml)

PR #193, opened 2026-09-08
Stop DedupConversationFilter from keeping retweets of mid-thread replies

Pitchfork-and-Torch:
DedupConversationFilter keys replies by `min(ancestors)` and ancestor-less retweets by `get_original_tweet_id()`. That only collapses a retweet of the root. A retweet of a mid-thread reply keys to the reply id, so the RT and the rest of the thread both survive post-selection.

Proof
- Entry: DedupConversationFilter::get_conversation_id (phoenix post-selection)
- Sink: For You TopK after VF; one kept post per conversation_id
- Break: RT of reply 11 with ancestors [2, 1] keys to 11; the reply keys to 1
- Viewer: a viral reply-RT and another post from the same thread both take For You slots
- Twin: existing tests already collapse RT-of-root (retweeted_tweet_id == min(ancestors)); Thunder fills ancestors on replies, not on retweets

This is not #97/#102 (gap expansion / unhydrated grandparents). Not #99 (self-reply chains). Not #101/#111 (muted/blocked ancestors). Not #179 (seen/served parent). Not #180 (Phoenix VF parents).

Fix: map each slate tweet and ancestor onto the smallest conversation root already present. Ancestor-less retweets use that root when the original is on the slate as a reply or as someone's ancestor. Unrelated originals stay on their own tweet id. Replies still key by `min(ancestors)`, so shallow [2] vs [2, 1] stay two conversations.

One file on current main. Replaces contaminated #189 (19 files, behind main).

Tests: RT of the reply itself, RT of a reply listed as a sibling ancestor, unrelated RT unchanged, existing root-RT and multi-conversation cases unchanged.

cargo test cannot run here. Public dump has no Home Mixer manifest. Standalone rustc model of old vs new keying passed 5/5.

### DIFF
--- a/home-mixer/filters/dedup_conversation_filter.rs
+++ b/home-mixer/filters/dedup_conversation_filter.rs
@@ -11,13 +11,14 @@ impl Filter<ScoredPostsQuery, PostCandidate> for DedupConversationFilter {
         _query: &ScoredPostsQuery,
         candidates: Vec<PostCandidate>,
     ) -> FilterResult<PostCandidate> {
+        let tweet_to_root = conversation_roots(&candidates);
         let mut kept: Vec<PostCandidate> = Vec::with_capacity(candidates.len());
         let mut removed: Vec<PostCandidate> = Vec::new();
         let mut best_per_convo: FxHashMap<u64, (usize, f64)> =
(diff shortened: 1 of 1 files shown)


Pitchfork-and-Torch:

